PR #7, opened 2020-02-24
safer temp names

ocefpaf:
~~Not ready for review yet.~~

This is ready for review now.

### DIFF
--- a/conda_lock.py
+++ b/conda_lock.py
@@ -85,7 +85,7 @@ def make_lock_files(platforms, channels, specs):
 
     print("To use the generated lock files create a new environment:", file=sys.stderr)
     print("", file=sys.stderr)
-    print("     conda create -n YOURENV --file conda-linux-64.lock", file=sys.stderr)
+    print("     conda create --name YOURENV --file conda-linux-64.lock", file=sys.stderr)
     print("", file=sys.stderr)
 
 
--- a/conda_lock.py
+++ b/conda_lock.py
@@ -20,7 +20,6 @@ import tempfile
 
 
 DEFAULT_PLATFORMS = ["osx-64", "linux-64", "win-64"]
-FAKE_PKGS_ROOT = "/tmp/something/that/does/not/exist"
 
 
 def _safe_fake_prefix_name():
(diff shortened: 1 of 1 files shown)


conda_lock.py
```diff
@@ -17,19 +17,28 @@
...
-FAKE_PKGS_ROOT = "/tmp/something/that/does/not/exist"
 
 
+def _safe_fake_prefix_name():
+    args = ("conda", "env", "list", "--quiet", "--json")
+    json_output = subprocess.check_output(args)
+    envs = json.loads(json_output)["envs"]
+    PREFIX_NAMES = [pathlib.Path(env).stem for env in envs if "envs" in env]
+
+    while (FAKE_PREFIX_NAME := next(tempfile._get_candidate_names())) not in PREFIX_NAMES:
```
ocefpaf: I don't like this b/c it will limit to Pyhton 3.8 and uses a private method.

mariusvniekerk: basically we just need a way to get a thing to find a writable prefix (which is somehow required for conda to solve even in dry-run mode)

mariusvniekerk: So we can just make a new tempdir and pick a random prefix inside there

conda_lock.py
```diff
@@ -16,31 +16,33 @@
...
-    json_output = subprocess.check_output(args, env=env)
-    # print(json_output)
+    with tempfile.TemporaryDirectory() as CONDA_PKGS_DIRS:
+        env.update({"CONDA_SUBDIR": platform, "CONDA_PKGS_DIRS": CONDA_PKGS_DIRS})
+
+        args = [
+            "conda",
+            "create",
+            "--prefix",
+            f"{CONDA_PKGS_DIRS}_prefix",
```
ocefpaf: @mariusvniekerk, after asking around I believe this is safer b/c, if one day dry-run does the right thing and fails with an existing name, we would have a broken conda-lock.

conda_lock.py
```diff
@@ -16,31 +16,33 @@
...
-        "--override-channels",
-        "--dry-run",
-        "--json",
-    ]
-    for channel in channels:
-        args.extend(["-c", channel])
-    args.extend(specs)
 
     env = dict(os.environ)
-    env.update({"CONDA_SUBDIR": platform, "CONDA_PKGS_DIRS": FAKE_PKGS_ROOT})
```
ocefpaf: Changed the names of the variables to match the conda equivalent for readability. I also expanded all the short optional args for the same reason.

conda_lock.py
mariusvniekerk:
```suggestion
            f"{CONDA_PKGS_DIRS}/prefix",
```

mariusvniekerk: or maybe with pathlib's join

Issue #12, opened 2020-02-27
KeyError: 'ca-certificates-2019.11.28-hecc5488_0'


rsignell-usgs:
The following environment.yml fails to lock:
```yaml
name: pangeo
channels:
  - conda-forge
dependencies:
  - python
  - bokeh
  - cartopy
  - cython
  - dask
  - dask-cloudprovider
  - erddapy
  - numpy
  - python-blosc
  - hvplot
  - h5py
  - h5netcdf
  - ipykernel
  - ipywidgets
  - geoviews
  - holoviews
  - lz4
  - datashader
  - panel
  - metpy
  - netcdf4
  - xarray
  - s3fs
  - gcsfs
  - fsspec
  - utide
  - zarr
  - earthsim
```

```
sh-4.2$ conda-lock -f pangeo_env.yml -p linux-64
generating lockfile for linux-64
Traceback (most recent call last):
  File "/home/ec2-user/anaconda3/bin/conda-lock", line 8, in <module>
    sys.exit(main())
  File "/home/ec2-user/anaconda3/lib/python3.6/site-packages/conda_lock/conda_lock.py", line 124, in main
    platforms=args.platform or DEFAULT_PLATFORMS,
  File "/home/ec2-user/anaconda3/lib/python3.6/site-packages/conda_lock/conda_lock.py", line 84, in make_lock_files
    url = urls[pkg["dist_name"]]
KeyError: 'ca-certificates-2019.11.28-hecc5488_0'
sh-4.2$
```

mariusvniekerk:
I managed to lock that environment just fine.

https://github.com/mariusvniekerk/conda-lock/blob/master/example/pangeo/conda-linux-64.lock

Can you share your `.condarc` and the conda version used to run the lock?

rsignell-usgs:
my `.condarc`:
```
(base) sh-4.2$ more ~/.condarc
show_channel_urls: true
channel_priority: strict
channels:
  - conda-forge
envs_dirs:
  - /home/ec2-user/SageMaker/myenvs
pkgs_dirs:
  - ~/anaconda3/pkgs
auto_update_conda: false
safety_checks: disabled
```
conda version
```
(base) sh-4.2$ conda --version
conda 4.8.2
```

Today's error: 
```
(base) sh-4.2$ conda-lock -f pangeo_env.yml -p linux-64
generating lockfile for linux-64
Traceback (most recent call last):
[...]

PR #13, opened 2020-02-29
Allow conda-lock to work when pkgs_dir already contains artifacts

mariusvniekerk:
This is a workaround for the issue that manifests in #12 .

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -238,7 +238,7 @@ def make_lock_files(
 
 def main_on_docker(env_file, platforms):
     env_path = pathlib.Path(env_file)
-    platform_arg = [] 
+    platform_arg = []
     for p in platforms:
         platform_arg.extend(["--platform", p])
 
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -182,13 +182,13 @@ def fn_to_dist_name(fn: str) -> str:
 def make_lock_files(
     conda: PathLike, platforms: List[str], channels: List[str], specs: List[str]
 ):
-    for platform in platforms:
-        print(f"generating lockfile for {platform}", file=sys.stderr)
+    for plat in platforms:
+        print(f"generating lockfile for {plat}", file=sys.stderr)
         dry_run_install = solve_specs_for_arch(
-            conda=conda, platform=platform, channels=channels, specs=specs
+            conda=conda, platform=plat, channels=channels, specs=specs
         )
-        with open(f"conda-{platform}.lock", "w") as fo:
-            fo.write(f"# platform: {platform}\n")
+        with open(f"conda-{plat}.lock", "w") as fo:
+            fo.write(f"# platform: {plat}\n")
             fo.write("@EXPLICIT\n")
             link_actions = dry_run_install["actions"]["LINK"]
             for link in link_actions:
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -238,7 +238,7 @@ def make_lock_files(
 
 def main_on_docker(env_file, platforms):
     env_path = pathlib.Path(env_file)
-    platform_arg = list()
+    platform_arg = [] 
     for p in platforms:
         platform_arg.extend(["--platform", p])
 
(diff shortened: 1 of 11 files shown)


conda_lock/conda_lock.py
```diff
@@ -23,42 +31,130 @@
...
+        return conda_executable
+
+    conda_executable = shutil.which("conda.exe")
+    if conda_executable:
+        return conda_executable
+
+    logging.info(
+        "No existing conda installation found.  Installing the standalone conda solver"
+    )
+    return install_conda_exe()
```
ocefpaf: Nice to check if conda exists but do we really want to install it for the use?

conda_lock/conda_lock.py
```diff
@@ -23,42 +31,130 @@
...
+
+
+def install_conda_exe():
+    conda_exe_prefix = "https://repo.anaconda.com/pkgs/misc/conda-execs"
```
ocefpaf: Are those advertised and/or recommended for use? That is a great idea but I'm worried if they are well maintained.

conda_lock/conda_lock.py
```diff
@@ -23,42 +31,130 @@
...
+        fo.write(resp.content)
+    st = os.stat(target_filename)
+    os.chmod(target_filename, st.st_mode | stat.S_IXUSR)
+    return target_filename
+
+
+CONDA_PKGS_DIRS = None
+
+
+def conda_pkgs_dir():
```
ocefpaf: Does it make sense to upstream that to `conda` itself? Smells like a workaround to something that should be in conda.

conda_lock/conda_lock.py
```diff
@@ -23,42 +31,130 @@
...
+
+
+def solve_specs_for_arch(
+    conda: PathLike, channels: List[str], specs: List[str], platform: str
+) -> dict:
+    args: MutableSequence[PathLike] = [
+        conda,
+        "create",
+        "--prefix",
+        str(pathlib.Path(conda_pkgs_dir()).joinpath("prefix")),
```
ocefpaf: With the `PathLike` type we can remove the `str()` here, right?
(I added that and it is quite ugly, sorry.)

conda_lock/conda_lock.py
mariusvniekerk: Talked with jjhelmus and they are released and maintained in the same cadence as Anaconda distribution, so it should be fine

ocefpaf: That is great news!

conda_lock/conda_lock.py
mariusvniekerk: There is an open issue on conda about this already https://github.com/conda/conda/issues/6902

conda_lock/conda_lock.py
mariusvniekerk:
```suggestion
        pathlib.Path(conda_pkgs_dir()).joinpath("prefix"),
```

conda_lock/conda_lock.py
```diff
@@ -23,42 +31,130 @@
...
+        "--dry-run",
+        "--json",
+    ]
+    for channel in channels:
+        args.extend(["--channel", channel])
+    args.extend(specs)
+
+    try:
+        proc = subprocess.run(
+            args, env=conda_env_override(), capture_output=True, encoding="utf8"
```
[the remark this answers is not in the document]
mariusvniekerk: Technically we don't but that's mostly just because `json.loads` accepts input it probably shouldn't.  Under pypy this would be required

conda_lock/conda_lock.py
ocefpaf: 2018!

conda_lock/conda_lock.py
mariusvniekerk: Since we're just placing it in the installed path of our package, we're not so much installing as downloading optional package data?

conda_lock/conda_lock.py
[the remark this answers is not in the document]
ocefpaf: > Under pypy this would be required.

Also python 3.5, so let's keep it then.

conda_lock/conda_lock.py
ocefpaf: OK.

PR #15, opened 2020-03-02
Fix ensure conda

ocefpaf:
@mariusvniekerk I can squash the commits later but left them here to make it easier to review. The failures and fixes are tracked in the logs here: https://travis-ci.com/ocefpaf/conda-lock/builds


setup.cfg
```diff
@@ -30,7 +29,7 @@ zip_safe = True
 install_requires =
     pyyaml
     requests
-python_requires = >=3.6
```
mariusvniekerk: Why the python version change here?

ocefpaf: B/c I was lazy and did not want to fix this for 3.6:

https://travis-ci.com/ocefpaf/conda-lock/builds/151423694

ocefpaf: The thing is, we never supported Python 3.6 b/c of the `capture_output` in the `subprocess.run`, that was added in 3.7. Do you want to remove that and support 3.6 or let's keep it >=3.7?

mariusvniekerk: i'd probably say we should keep 3.6 compat, its still in pretty heavy use

ocefpaf: Done!

Issue #4, opened 2020-02-24
Support pip packages and pip GH installs in a conda environment.yaml file list.


ocefpaf:
If you have something like:

```yaml
name: test
channels:
  - conda-forge
dependencies:
  - python=3.7
  - pip:
    - some-pkg
    - git+https://github.com/someuser/another-pkg.git@master
```

conda-lock will crash. We should either fail more gracefully or add support pip installs too.

marcelotrevisani:
+1 for support of ``pip install --no-deps``

mariusvniekerk:
main issue with pip is lack of determinism.  But we can probably work around that

ocefpaf:
The install from `master` example above is just wrong IMO and I added it to discuss some sort of warning for conda-lock. For PyPI packages we can probably look into pip-tools for inspiration.

ocefpaf:
> +1 for support of `pip install --no-deps`

In a way that is what conda-env does, right? Not sure if it issues the --no-deps though.

I don't want to make this overly complicated, the first pass should be just a warning instead of a failure: "you have pip packages in your, this env won't pass a round trip and won't be fully reproducible. The pip packages are dropped from the lock file."

noahp:
Pip can check hashes for downloaded files:
https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode

ocefpaf:
> Pip can check hashes for downloaded files:
> [pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode](https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode)

I guess that the main challenge here is to make conda really understand what was installed by `pip`. I'm not even sure that an old bug, that made packages with an `-` in the name, be wrongly identified as `pip` packages even when installed with `conda`.

TL;DR this may be more difficult than it sounds b/c one would have to dig into `conda` itself.

@mariusvniekerk what do you say? Am I way off here?

mariusvniekerk:
So we can put some of the pip packages as a magic comment.  
This also does imply that conda-lock have an install mode so that it runs the appropriate pip installs since I'm pretty sure conda install isn't going to

sterlinm:
[...]

PR #25, opened 2020-05-24
Add 'lock' files for pip dependencies

noahp:
This attempts to address some parts of #4:

- no longer error when producing lock files for environments with pip
dependencies
- output a `*.lock.pip` file that contains the pip-specified
dependencies for the target environment

Note that if any pip args were passed in the original environment spec
file, they're dropped, since conda-lock processes the resolved
environment from conda (eg if you pass `--extra-index-url` etc).

Included some help text when producing pip 'lock' files.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -178,8 +178,31 @@ def parse_environment_file(environment_file: pathlib.Path) -> Dict:
     # TODO: we basically ignore most of the fields for now.
     #       notable pip deps are not supported
     specs = env_yaml_data["dependencies"]
+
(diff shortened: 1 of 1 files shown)


noahp:
Hmm, this is actual probably less useful than the user just running `conda update` with the original environment spec file :confused:

munro:
@noahp any reason why you closed this?  my `environment.yml`'s pip dependencies don't have any versions—I like to just update everything, and fix any breakage all at once.  so locking pip dependencies would be very nice when I'm not in the mood for an upgrade.

munro:
Ahh I see this being tracked in #4

PR #28, opened 2020-06-03
Add better tests and error reporting

mariusvniekerk:
cc @ocefpaf 

Also added tests that ensure that running with mamba works


.travis.yml
```diff
@@ -35,6 +35,7 @@ before_install:
     conda config --add channels conda-forge --force
     conda config --set channel_priority strict
     conda config --set safety_checks disabled
+    conda install mamba
```
ocefpaf, line 38: Let's add it to the requirements-dev.txt so people don't have to dig into our CIs to run the tests.

conda_lock/conda_lock.py
```diff
@@ -32,26 +33,27 @@
...
 
-def ensure_conda(conda_executable: Optional[str] = None):
+def candidate_executables(conda_executable: Optional[str]) -> Iterable[Optional[str]]:
     if conda_executable:
         if pathlib.Path(conda_executable).exists():
-            return conda_executable
+            yield conda_executable
+        yield shutil.which(conda_executable)
+    yield shutil.which("conda")
+    yield shutil.which("conda.exe")
```
ocefpaf, line 42: Nice!

conda_lock/conda_lock.py
```diff
@@ -256,8 +269,8 @@ def make_lock_files(
             for pkg in link_actions:
                 url = fetch_by_dist_name[pkg["dist_name"]]["url"]
                 md5 = fetch_by_dist_name[pkg["dist_name"]]["md5"]
-                r = requests.head(url, allow_redirects=True)
-                url = r.url
+                # r = requests.head(url, allow_redirects=True)
```
ocefpaf, line 272: This will make it faster and we don't need to check the URLs. Maybe just remove these lines?

.travis.yml
mariusvniekerk: Mostly since mamba is not pip installable atm, so want to not force that, since it will skip that test if mamba isn't around

ocefpaf: I always forget about pip ;-p

But do we care about that in a tool that is all about conda?

PR #32, opened 2020-07-15
Add first class mamba support

mariusvniekerk:
When mamba is available, prefer it over stock conda


ocefpaf:
First I thought that we should have a config/option for that. However, if someone installed mamba already this is what they want. Merge?

PR #33, opened 2020-07-15
Add support for creating a lock file from a meta.yaml

### LINKED ISSUE #31: Allow creating a lockfile from a meta.yaml

mariusvniekerk:
Added support for creating a lockfile by pointing at a meta.yaml file.

This uses a minimal meta.yaml parser that is platform aware (but nothing else) to generate the requirements

Resolves #31

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -19,7 +19,10 @@ import tempfile
 from typing import Dict, Iterable, List, MutableSequence, Optional, Set, Tuple, Union
 
 import requests
-import yaml
+
+from conda_lock.src_parser import LockSpecification
+from conda_lock.src_parser.environment_yaml import parse_environment_file
+from conda_lock.src_parser.meta_yaml import parse_meta_yaml_file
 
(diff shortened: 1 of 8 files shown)


ocefpaf:
LGTM. Maybe @marcelotrevisani have some advice on the meta.yaml parsing b/c I know he has been working on that for a while.

marcelotrevisani:
I am using ruamel instead of pyyaml because ``pyyaml`` does not handle well comments and it does not support much configurations.
One thing which I am thinking to do is to extract the code responsible to handle the recipe from grayskull to a separated project, and there we can be very specific to parse conda`s recipes

scopatz:
A brief look and things look good!  I don't think comments much matter here.  Conda-lock won't ever try to write the YAML back out.  The only place I  can see it being an issue is when selectors are behind comments

ocefpaf:
@mariusvniekerk, new release?

mariusvniekerk:
Yeah a new release sounds good 0.4.0

PR #35, opened 2020-08-11
Improved error reporting when errors aren't reported in the json response

mariusvniekerk:


conda_lock/conda_lock.py
```diff
@@ -189,6 +188,14 @@ def solve_specs_for_arch(
         stderr=subprocess.PIPE,
         encoding="utf8",
     )
+
+    def print_proc(proc):
+        print(f"    Command: {proc.args}")
+        if proc.stdout:
+            print(f"    STDOUT:\n{proc.stdout}")
+        if proc.stderr:
+            print(f"    STDOUT:\n{proc.stderr}")
```
marcelotrevisani, line 197: Should it be ``STDERR`` then?
```python
print(f"    STDERR:\n{proc.stderr}")
```

PR #36, opened 2020-08-11
Harden up undefined vars jinja support a bit

mariusvniekerk:
Harvested the super undefined helper from conda


mariusvniekerk:
yeah this doesn't quite work

PR #37, opened 2020-08-11
Fix print message for stderr

marcelotrevisani:


marcelotrevisani:
related to https://github.com/mariusvniekerk/conda-lock/pull/35#discussion_r468635985

ocefpaf:
Thanks for fixing my bad review!

marcelotrevisani:
> Thanks for fixing my bad review!

not at all, those things are hard to catch :smile:

PR #38, opened 2020-08-14
Add support for creating a conda-lock from a pyproject.toml

mariusvniekerk:
Building on some of the work in grayskull we can actually generate a conda-lock from
the package specification in a poetry pyproject.toml.

Support for flit/setup.cfg/Pipfile should be fairly easy to add in the future

### DIFF
--- a/conda_lock/src_parser/poetry.py
+++ b/conda_lock/src_parser/poetry.py
@@ -1,0 +1,98 @@
+import collections
+import collections.abc
+import copy
+import pathlib
+
+from typing import List, Optional
+
+import requests
+import toml
+import yaml
+
+from conda_lock.src_parser import LockSpecification
+
+
+# TODO: make this configurable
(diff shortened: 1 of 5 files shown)


conda_lock/src_parser/poetry.py
```diff
@@ -0,0 +1,98 @@
...
+
+import requests
+import toml
+import yaml
+
+from conda_lock.src_parser import LockSpecification
+
+
+# TODO: make this configurable
+PYPI_TO_CONDA_NAME_LOOKUP = "https://raw.githubusercontent.com/marcelotrevisani/grayskull/master/grayskull/pypi/config.yaml"
```
ocefpaf, line 16: We need a stable place for this in conda-forge. Grayskull should read from that. A simple repo with that file would be useful in many projects.

mariusvniekerk: yeah we're planning on having the bot autogenerate this

scopatz:
This LGTM! Should I merge?

PR #49, opened 2020-09-10
Add ability to specify channel overrides on the command line

mariusvniekerk:


conda_lock/conda_lock.py
```diff
@@ -280,10 +280,13 @@ def make_lock_files(
             include_dev_dependencies=include_dev_dependencies,
         )
 
+        if not override_channels:
+            channels = lock_spec.channels
```
ericdill, line 284: you probably dont need this extra flag. I think you implicitly have this flag with `channels != ()`

ericdill: i.e., if any channels are provided to this function then you're overriding what's in the lock spec

conda_lock/conda_lock.py
```diff
@@ -372,6 +375,23 @@ def parser():
...
+        """,
+        default=False,
+    )
+    parser.add_argument(
+        "-c",
+        "--channel",
+        nargs="?",
+        action="append",
+        help="Additional channels to include in the solve.  Requires the --override-channels flag",
+    )
```
ericdill, line 393: I think you need a default value here too or args.channels is gonna raise down on line 469

mariusvniekerk: changed the channels to allow None

conda_lock/conda_lock.py
```diff
@@ -442,6 +481,8 @@ def main():
         platforms=args.platform,
         no_mamba=args.no_mamba,
         include_dev_dependencies=args.dev_dependencies,
+        channels=args.channels,
```
ericdill, line 484: delete this

ericdill (approved):

PR #53, opened 2020-09-17
Pass str to subprocess calls

kev-zheng:
https://github.com/conda-incubator/conda-lock/issues/52

With this change, generates successfully:
image


scopatz:
Cool Thanks!

PR #54, opened 2020-09-18
cast conda exec to string

kev-zheng:
For some reason I missed this... or the conda executable sometimes just spuriously becomes a string / becomes a WindowsPath.

Here the conda.BAT exec for windows is a WindowsPath:
image

with fix:
image


scopatz:
Thanks!

PR #56, opened 2020-10-07
Fix incorrect matchspec construction

mariusvniekerk:
This caused incorrect solves with poetry/flit based conda lock attempts


scopatz:
LGTM thanks!

PR #58, opened 2020-10-09
Update precommit action

mariusvniekerk:


scopatz:
Do we expect that precommit will fail?

mariusvniekerk:
guess i wrote the yaml bad so yeah?

PR #59, opened 2020-10-13
Add support for conda-lock install

### LINKED ISSUE #43: Add command to perform a conda install
```
conda-lock install [-p {foo}|-n {foo}] mylockfile.lock --additional args to conda
```

munali:
Closes #43

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,6 @@ matrix:
       env: PY=3.7
     - name: "python-3.8"
       env: PY=3.8
-    - name: "coding_standards"
-      env: PY=3
-    - name: "tarball"
-      env: PY=3
     - name: "docs"
       env: PY=3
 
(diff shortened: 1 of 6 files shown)


tests/test_conda_lock.py
```diff
@@ -177,3 +184,164 @@ def test_aggregate_lock_specs():
...
+    return True
+
+
+@pytest.fixture
+def conda_exe():
+    return determine_conda_executable("conda", no_mamba=True)
+
+
+@pytest.fixture
+def conda_env(conda_exe):
```
mariusvniekerk, line 272: I don't think we need this fixture, since this tests that we can install things in a way we're not using?

mariusvniekerk: we can use the destroy with a finally in the test

conda_lock/conda_lock.py
```diff
@@ -121,6 +123,58 @@ def print_proc(proc):
...
+        if proc.stdout:
+            print(f"    STDOUT:\n{proc.stdout}")
+        if proc.stderr:
+            print(f"    STDERR:\n{proc.stderr}")
+
+    try:
+        proc.check_returncode()
+    except subprocess.CalledProcessError:
+        try:
+            err_json = json.loads(proc.stdout)
```
mariusvniekerk, line 164: this will only work if there is a `--json` as part of the conda create command

mariusvniekerk (approved):

.travis.yml
```diff
@@ -17,10 +17,6 @@ matrix:
       env: PY=3.7
     - name: "python-3.8"
       env: PY=3.8
-    - name: "coding_standards"
-      env: PY=3
-    - name: "tarball"
```
ocefpaf, line 22: Why remove these tests?

.travis.yml
```diff
@@ -17,10 +17,6 @@ matrix:
       env: PY=3.7
     - name: "python-3.8"
       env: PY=3.8
-    - name: "coding_standards"
```
ocefpaf, line 20: This one can go now that we have pre-commit hooks that do the same.

ocefpaf (approved):

.travis.yml
ocefpaf: The tarball test do help prevent bad releases. I recommend to add them back.

PR #60, opened 2020-10-23
Add crude parsers for multioutput meta.yaml

mariusvniekerk:
For simplicity we just assume you want everything in all outputs


ocefpaf (approved):
Makes sense to me.

PR #62, opened 2020-10-27
add micromamba usage

munali:


conda_lock/conda_lock.py
```diff
@@ -372,23 +372,31 @@ def aggregate_lock_specs(lock_specs: List[LockSpecification]) -> LockSpecificati
...
-        # micromamba doesn't support --override-channels
-        micromamba=False,
+        mamba=mamba,
+        micromamba=micromamba,
         conda=True,
         conda_exe=True,
     )
+
+    if micromamba and "MAMBA_ROOT_PREFIX" not in os.environ:
+        os.environ["MAMBA_ROOT_PREFIX"] = str(pathlib.Path(_conda_exe) / "mamba_root")
```
mariusvniekerk, line 391:
```suggestion
        os.environ["MAMBA_ROOT_PREFIX"] = str(pathlib.Path(_conda_exe).parent / "mamba_root")
```

conda_lock/conda_lock.py
```diff
@@ -372,23 +372,31 @@ def aggregate_lock_specs(lock_specs: List[LockSpecification]) -> LockSpecificati
...
     if conda_executable:
         if pathlib.Path(conda_executable).exists():
             yield conda_executable
         yield shutil.which(conda_executable)
+
     _conda_exe = ensureconda.ensureconda(
-        mamba=not no_mamba,
-        # micromamba doesn't support --override-channels
-        micromamba=False,
+        mamba=mamba,
```
mariusvniekerk, line 384: do we have a test that runs a lock using micromamba?

conda_lock/conda_lock.py
munali: good catch!

conda_lock/conda_lock.py
munali: no, I will add a test.

mariusvniekerk:
Nice.   Once micromamba support CONDA_SUBDIR we can remove the xfail and see about getting it enabled by default.

PR #64, opened 2020-11-13
diagnose test failure

mariusvniekerk:


.github/workflows/test.yml
```diff
@@ -12,48 +12,56 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
-          auto-activate-base: true
           channels: conda-forge
 
       - name: Install deps
-        shell: bash -l {0}
         run: |
-          conda install mamba pip
-          conda activate base
+          conda activate test
```
goanpeca, line 29: @mariusvniekerk we do this by default for you: so you can have 3 cases:

Either you use `base` by default, in which case you need

## Activate `base` for all steps

```yaml
 - uses: conda-incubator/setup-miniconda@v2
        with:
          auto-update-conda: true
          python-version: ${{ matrix.python-version }}
          auto-activate-base: true	
          activate-environment: ""
          channels: conda-forge
```

## Activate `whatever` for all steps

```yaml
 - uses: conda-incubator/setup-miniconda@v2
        with:
          auto-update-conda: true
          python-version: ${{ matrix.python-version }}
          auto-activate-base: true
          activate-environment: whatever
          channels: conda-forge
```


## Activate `test` for all steps

```yaml
 - uses: conda-incubator/setup-miniconda@v2
        with:
          auto-update-conda: true
          python-version: ${{ matrix.python-version }}
          auto-activate-base: true
          channels: conda-forge
```

PR #75, opened 2020-12-31
Logo

mariusvniekerk:
This likely needs a bit more work and some attribution


bollwyvl:
Ooooh, icon time!

Here's another:

ocefpaf:
@bollwyvl is it me or that looks like a suit case with the a handle to the side instead of the center?

ocefpaf:
> This likely needs a bit more work and some attribution

We should bring this up in the conda community meeting to get the blessing but I really like this one!

bollwyvl:
> looks like a suit case with the a handle to the side

Is that so inaccurate? Suit cases have locks. You put your... important business papers in them. 
Or an MC-Escherian coffee cup, according to my in-house design team? But the suggested `CL` is the important part, I suppose. I'll do some alternates.

Aside: there's no suitcase emoji only :business_suit_levitating: which is kinda creepy.

bollwyvl:
Here's more of a kettlebell, i guess:

bollwyvl:
There's probably something that can be done to make the `C` more directly look like a keyhole...

ocefpaf:
> But the suggested `CL` is the important part

Oh! I missed that. Now I can see it :-)
I'm still on the first one though.

bollwyvl:
Yeah, the OG conda logo has always struck me as a little hard to use as a design asset... happy to do anything that doesn't propagate it further :P

janjagusch:
what about something that resembles a key?

image
image

mithro:
Your logo made me think you could make it into a snake... Here is my artistically challenged attempt :-)

image

bollwyvl:
Love the ideas... but that particular logo is a registered trademark, we shouldn't use it (even if it did fit within the usage guidance, which i haven't, and probably won't, check).

the derivative (or maybe it's reversed) conda logo, as mentioned above, is also pretty hard to use, as it's too detailed for small logos.

wagnerpeer:
Came across this PR and would like to put my two cents in. Especially liked the idea of the kettlebell/padlock (modified version of logo by @bollwyvl) How to come to a conclusion? Is there a formal way?
conda-lock-v1

Issue #69, opened 2020-12-04
Output options: file name, standard out?


bollwyvl:
Thanks again for `conda-lock`!

Two use cases:

- I keep finding myself generating a lock file in a temporary directory and then renaming it.
- I also occasionally want a single lockfile, not even on disk e.g. verify that a particular channel is not used.

The potentially-multi-platform output of course complicates both of these use cases... practically, I just about always end up with extra windows packages/channels, so would be totally fine with `--output` or `--stdout` being mutually exclusive with multi-platform outputs.

Something that _could_ work with multiple outputs would be a prefix: looking through the code, there's a spot where the filename is generated from `f"conda-{platform}.lock"`... and knowing the name of the platform is of course pretty important! But perhaps being able to specify a prefix, e.g. `f"{output_prefix}conda-{platform}.lock"` would keep things consistent. If anything, I'd almost prefer a composite extension, e.g. `.conda.lock`, but probably that ship has sailed.

Supporting e.g. `--stdout`, even for the single case, is a bit more challenging, as there is a fair amount of `print` stuff, but this could be overcome with `click.echo(..., err=True)`, which does some nice stuff.

Anyhow, architecturally, this might mean reworking `make_lock_files` a bit.

Happy to work up a PR if this doesn't sound like crazy talk!

habemus-papadum:
I was wanting this too.  Is there any time savings by solving multiple platforms in one invocation?

bollwyvl:
I doubt there's much shared, as it shells out to `$CONDA_EXE`...

ma-sadeghi:
+1

mariusvniekerk:
When solving multiple platforms at once we would still need to download `n` platforms worth of repodata, so any gains are going to be very marginal.

@bollwyvl That sounds like a good plan

For the filenames we should allow you to specify the filename on the cli as `"something{platform}something"`, with a default for the current template of `"conda-{platform}.lock"`

bollwyvl:
I like the template idea better than anything else I've come up with. Maybe
make the env hash an available token, etc...

mariusvniekerk:
also allows us to expand the available set of things easily.  dev-dependencies is also a good token.

Not sure if we want just a simple `.format()` or something heavier like jinja.  Leaning towards the former for simplicity

mariusvniekerk:
Most of the ideas are now implemented in v0.8.0. 
@bollwyvl feel free to close this if it suits your needs now

[...]

PR #78, opened 2021-03-03
Only require {platform} in template if building for multiple platforms

minrk:
allows single-platform locks to specify exact filenames

Related to #69.

Additionally, moves the argument validation earlier so it raises before running any locks, instead of after the first run.


ocefpaf (approved):

ocefpaf:
@mariusvniekerk I think this is a good idea, if you agree do you mind merging it? I can take care of a new release.

PR #95, opened 2021-05-21
Add selectors that match osx-arm64 (arm64, osx, unix)

ryanvolz:
Currently, the use of any selectors fails when creating a lockfile for osx-arm64 (or any platform that's not listed here). Adding osx-arm64 selectors is the simplest fix, although a more general solution that gracefully handles new platforms would obviously be desirable. At least this brings the list up to parity for what conda-forge currently builds for.


mariusvniekerk:
Thanks for the submission

PR #102, opened 2021-08-06
Sort lockfile entries by package name

acarapetis:
First of all, thanks for providing this great tool! Combined with micromamba it has made deploying python apps a much nicer experience.

One small bother I've had is the non-deterministic ordering of packages in the lockfiles produced by conda-lock. If I modify my requirements and then re-run conda-lock, I'd like to be able to quickly glance at the diff as a sanity check; but even if only one package has actually changed, the re-ordering will make the diff huge.

This PR solves this by sorting the packages by name before outputting them to the lockfile.

My understanding of conda is very limited, so I'm not 100% sure the order of packages in the lockfile isn't important - please enlighten me and reject the PR if it is!

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -6,6 +6,7 @@ import atexit
 import datetime
 import json
 import logging
+from operator import itemgetter
 import os
 import pathlib
 import re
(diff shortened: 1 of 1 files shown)


mariusvniekerk:
The order of entries in the lockfile is important since that ensures that packages are installed in topological order.  

This matters in some cases where packages clobber each other's files and later ones take precendence as well as for packages that have post-link step.

PR #107, opened 2021-09-08
Fix regression for multi-platform lock

Japanuspus:
Regression was  introduced in 9aa90dc0b12c49785e912a1f87bfe38762e89058: `make_lock_specs` would write all resolved specs to the key `sys.platform`.
This results in only the first platform being processed. There are no error messages and only hint of something being amiss is that logging reports the lockfile being built for the value of `sys.platform` instead of the actual platform.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -373,7 +373,7 @@ def make_lock_specs(
         else:
             channels = lock_spec.channels
         lock_spec.channels = channels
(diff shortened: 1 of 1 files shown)


Japanuspus:
Added an issue for this as #108

mariusvniekerk:
Thanks!

mariusvniekerk:
Thanks!

PR #117, opened 2021-10-27
Allow for disabling platform validation

SimeonStoykovQC:
Тhere was no way to set this flag to `False` and disable platform validation. Here is a link to [click docs about boolean flags](https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags).

Creating a fix PR seemed faster than describing the issue. Let me know if I am missing something. Also feel free to contribute to this PR (e.g. tests).


mariusvniekerk:
there were some changes in the behavior of micromamba so had to do some work to work around those

PR #125, opened 2021-11-17
Fix LockSpecification __str__

ThomasBlauthQC:
This fixes a small bug in the __str__ function of LockSpecification.


mariusvniekerk:
With the large rework this is no longer relevant

PR #127, opened 2021-11-22
Allow to pass different credentials for different channels in same domain

janjagusch:
When we first implement the feature to strip auths from lock files and later provide auth during the installation, we didn't consider the case, where different channels in the same domain require different auths.

Example:

`http://mydomain.org/mychannel1` requires the following auth: `username1:password1`
`http://mydomain.org/mychannel2` requires the following auth: `username2:password2`

This is something you weren't able to express until now because you were limited to providing one auth per domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,15 @@ In order to `conda-lock install` a lock file with its basic auth credentials str
 }
 ```
 
+If you have multiple channels that require different authentication within the same domain, you can additionally specify the channel like this:
+
+```json
+{
+  "domain.org/channel1": "username1:password1",
+  "domain.org/channel2": "username2:password2"
+}
+```
+
 You can provide the authentication either as string through `--auth` or as a filepath through `--auth-file`.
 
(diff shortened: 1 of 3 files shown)


janjagusch:
@mariusvniekerk, could you maybe take a look at this? this would be super useful to us. 🙂

janjagusch:
> @mariusvniekerk, could you maybe take a look at this? this would be super useful to us. 🙂

ping, @mariusvniekerk. 🙂

mariusvniekerk:
Can you expand the documentation section in the README a bit to show this in action?

janjagusch:
> Can you expand the documentation section in the README a bit to show this in action?

Sure, I will tackle that tomorrow. 🙂

janjagusch:
> Can you expand the documentation section in the README a bit to show this in action?

Sorry, took a bit longer than expected! Hope it's good to go now. 🙂

PR #131, opened 2021-12-06
Minor lockfile improvements

jvansanten:
Address points raised by @maresb and @wolfv in follow-on discussion in https://github.com/mamba-org/mamba/issues/1209:

- Include a usage blurb in conda-lock.yml to ward off attempts to edit the lock by hand
- Ensure package hash explicitly stores the hash algorithm used


maresb:
Wow @jvansanten, this is wonderful! Thanks so much for integrating my feedback about making the lockfile self-documenting. I think this will be extremely helpful. :pray:

tests/test-lockfile/conda-lock.yml
```diff
@@ -47,7 +62,7 @@ package:
   dependencies:
     python: '>=3.9,<3.10.0a0 *_cpython'
     python_abi: 3.9.* *_cp39
-  hash: 7fb8a37f9e9afaad0210974e1452f84a
+  hash: md5:7fb8a37f9e9afaad0210974e1452f84a
```
wolfv, line 65: is `md5:xyz` valid yaml or not? that's my only concern here.

wolfv: > YAML’s block collections use indentation for scope and begin each entry on its own line. Block sequences indicate each entry with a dash and space (“- ”). Mappings use a colon and space (“: ”) to mark each key/value pair. Comments begin with an octothorpe (also called a “hash”, “sharp”, “pound” or “number sign” - “#”).

Looks like it's not a map as there would need to be a space after the `md5`. I think it's fine since these files aren't meant to be edited by humans.

PR #147, opened 2022-02-15
Refactor lookup

ocefpaf:
I found an odd bug when using the latest dev version with pip support. If one have an environment with pip dependencies that installs a conda dependency with a name that doesn't match PyPI, like:

```yaml
name: test
channels:
  - conda-forge
dependencies:
  - python=3.10
  - dask
  - pip
  - pip: 
      - geogif  # brings in dask, that is in the env file too
```

The problem is that the error doesn't happen all the time and it is really frustrating trying to figure out why. Maybe the differences in the solver results from one run to another?. @mariusvniekerk, if you have some time I'd love to hear your thought on this.

Here is the error when it happens:

```
❯ conda-lock --conda=mamba -f environment.yaml -p linux-64
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "conda-lock/conda_lock/src_parser/__init__.py", line 251, in seperator_munge_get
    return d[key.replace("-", "_")]
KeyError: 'dask_core'

During handling of the above exception, another exception occurred:
```

This PR factors out some of the logic for the lookups so we can import them in multiple places avoiding circular imports. I needed that to use `normalize_conda_name` in multiple places. With that change the error never happens but I'm not confident this is the best solution.

### DIFF
--- a/conda_lock/src_parser/__init__.py
+++ b/conda_lock/src_parser/__init__.py
@@ -10,7 +10,7 @@ from typing import ClassVar, Dict, List, Literal, Optional, Sequence, Set, Tuple
 from pydantic import BaseModel, Field, validator
 
 from conda_lock.common import ordered_union
-from conda_lock.lookup import normalize_conda_name
+from conda_lock.lookup import conda_name_to_pypi_name, pypi_name_to_conda_name
 from conda_lock.virtual_package import FakeRepoData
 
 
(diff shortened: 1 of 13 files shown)


mariusvniekerk:
@ocefpaf its probably worth adding that example in as a test to ensure that we don't have future regressions.  This feels better than just the simple try a few underscore and dash styles.

mariusvniekerk:
This is also a much more reliable solve for issues like https://github.com/conda-incubator/conda-lock/issues/137

ocefpaf:
> @ocefpaf it is probably worth adding that example in as a test to ensure that we don't have future regressions. This feels better than just the simple try a few underscore and dash styles.

Done.

tests/test_conda_lock.py
```diff
@@ -378,6 +385,15 @@ def test_run_lock_with_pip(monkeypatch, pip_environment, conda_exe):
     run_lock([pip_environment], conda_exe=conda_exe)
 
 
+def test_run_lock_with_pip_environment_different_names_same_deps(
```
ocefpaf, line 388: Ouch! Sorry I missed that!!

conda_lock/src_parser/__init__.py
```diff
@@ -287,7 +278,7 @@ def seperator_munge_get(
 
     for dep, root in root_requests.items():
         source = requested[root]
-        target = seperator_munge_get(planned, dep)
+        target = seperator_munge_get(planned, normalize_conda_name(dep).lower())
```
mariusvniekerk, line 281: I think this place is too common to normalize the name of a dep here. We may need to update the dep model to have multiple names in it for pypi names and conda names

ocefpaf: Makes sense. Maybe we should close this one and start over? I have to move to another project but I can try to tackle that when I get more free time.

mariusvniekerk: Nah, i'll take this one over.  The core idea is right, just not in the correct spot.

mariusvniekerk:
Looks like it works now, just failing on some standard conda really hates concurrency cases.

Issue #148, opened 2022-02-16
Error in main when using blas[build=mkl] in environment.yml


lesteve:
Trying to use conda-lock with an environment.yml that has `blas[build=mkl]`
creates a `KeyError` on main. This seems to work fine with 0.13.2.

To reproduce:

```bash
cat << EOF > test-environment.yml
channels:
  - conda-forge

dependencies:
  - numpy
  - blas[build=mkl]
EOF
 
mamba create -n test-env conda-lock -y                       
conda activate test-env
pip install git+https://github.com/conda-incubator/conda-lock
conda-lock -p linux-64 -f test-environment.yml
```

```
❯ conda-lock lock -p linux-64 -f test-environment.yml 
<re.Match object; span=(0, 19), match='  - blas[build=mkl]'>
('  - blas', None, 'build=mkl')
Traceback (most recent call last):
  File "/home/lesteve/miniconda3/envs/conda-lock/bin/conda-lock", line 8, in <module>
    sys.exit(main())
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/conda_lock/conda_lock.py", line 1074, in lock
    lock_func(
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/conda_lock/conda_lock.py", line 873, in run_lock
    make_lock_files(
  File "/home/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/conda_lock/conda_lock.py", line 270, in make_lock_files
[...]

PR #150, opened 2022-02-16
For selector filtering require selectors to be commented

mariusvniekerk:
Conda meta.yaml selectors can take the form

```yaml
  - foo [linux]
```

instead of the more usual

```yaml
  - foo  # [linux]
```
which is basically entirely crazy.  Conda-lock will deviate from conda here by requiring selectors to be comments as they are in almost 100% of recipes


mariusvniekerk:
This should fix #148

PR #153, opened 2022-02-18
mkdocs

mariusvniekerk:
Switch docs to mkdocs and actually publish to gh-pages


jezdez:
Could I persuade you to use [MyST](https://myst-parser.readthedocs.io/en/latest/) instead? I think it would be cool to use Markdown of course, rST can die in a fire, but I was going to suggest that we'd add projects under the conda org to the docs.conda.io in the future using Read The Docs' umbrella project feature.

mariusvniekerk:
@jezdez Given how little of conda-lock (and really conda tooling in general) relies on autogenerated documentation from code which is the strong suite of sphinx mkdocs feels like a more natural documentation library for documentation that are almost entirely hand written.

I've based a fair bit of the documentation structure and template here on that from https://ibis-project.org/

And yes rST can go die in a fire.  

My experience with MyST is you still need to write a pretty large amount of stuff in rST and it adds a large number of strange failure conditions to running sphinx regularly.  

Even using something like https://bashtage.github.io/sphinx-material/ (unsure if it works with MyST) which gives a reasonably modern feeling documentation page it is still far behind the material theme for mkdocs in feature set. 

imho if you use sphinx, stay in rst land, life is better there, but accept the fact that your docs will look like they are stuck in the late 2000s.

jezdez:
> @jezdez Given how little of conda-lock (and really conda tooling in general) relies on autogenerated documentation from code which is the strong suite of sphinx mkdocs feels like a more natural documentation library for documentation that are almost entirely hand written.

Yeah, I consider that to be a bug that conda tooling isn't using more automated API documentation. I don't think it's utilized correctly at the moment, no idea why there wasn't a use of intersphinx more and a coordinated effort to align conda projects under one theme etc. All areas I'd like to spend some time in.

> I've based a fair bit of the documentation structure and template here on that from https://ibis-project.org/

Nicely looking docs!

> And yes rST can go die in a fire.
> 
> My experience with MyST is you still need to write a pretty large amount of stuff in rST and it adds a large number of strange failure conditions to running sphinx regularly.
> 
> Even using something like https://bashtage.github.io/sphinx-material/ (unsure if it works with MyST) which gives a reasonably modern feeling documentation page it is still far behind the material theme for mkdocs in feature set.

That's true, FWIW this wasn't an argument against mkdocs (I've used it successfully at Mozilla and have contributed to it), but an encouragement to make it easier to integrate conda documentation projects easier in the future.

> imho if you use sphinx, stay in rst land, life is better there, but accept the fact that your docs will look like they are stuck in the late 2000s.

Harsh, but probably fair. There are a few new themes and efforts that are useful to build a Sphinx site that aren't from the 2000s, so I haven't given up on Sphinx yet. I just don't see a similarly stable community around mkdocs though, the material theme for it is maintained by literally one person, something that I've tended to see as a huge risk long-term. The same is true for Sphinx and related themes though, so there is that.

Anyways, nothing is lost here to be clear, thanks for elaborating. When we get to working on the umbrella docs for the conda ecosystem we'll just need to make sure it'll allow non-Sphinx projects to be integrated well. Thanks!

PR #156, opened 2022-02-25
Expand channels to allow them to contain environment variables

mariusvniekerk:
These environment variables can get templated later at render time and are useful for things like usernames, passwords and tokens.

This enables the use-case where a channel is backed by Anaconda Enterprise or Quetz


mariusvniekerk:
@wolfv does micromamba not support channels with ports?

wolfv:
@mariusvniekerk do you have an example? Should be supported but there can always linger a bug!

mariusvniekerk:
The test failure here for micromamba is the example.

docs/authenticated_channels.md
```diff
@@ -0,0 +1,78 @@
...
+--channel 'http://host.com/t/$MY_REPO_TOKEN/channel'
+```
+
+If you accidentally pass a channel url that contains a token like so
+
+```sh
+--channel "http://host.com/t/$MY_REPO_TOKEN/channel"
+```
+
+Then conda lock will detect the environment variabl used (provided that the environment variable used ends in a sensible suffix (KEY, TOKEN, PASS, etc)).
```
cpcloud, line 24: This seems too magical. Is anything lost by not having this magic?

docs/authenticated_channels.md
```diff
@@ -0,0 +1,78 @@
...
+
+The _name_ of the environment variable will form part of your lock and you will have to have that SAME environment variable set if you wish to run the install
+
+```sh
+# retrieve secrets from some store
+source $(./get-lockfile-env-vars-from-secret-store)
+# use the secrets as part of the conda-lock invocation
+conda-lock install -n my-env-with-a-secret condae-lock.yml
```
cpcloud, line 32:
```suggestion
conda-lock install -n my-env-with-a-secret conda-lock.yml
```

docs/authenticated_channels.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Simple Auth
+
+For other channels (such as those self-managed) you may be using a more standard username/password auth
```
cpcloud, line 37:
```suggestion
For other channels (such as those self-managed) basic auth is supported
```

conda_lock/models/test_channel.py
```diff
@@ -0,0 +1,43 @@
...
+    monkeypatch.setenv("PASSWORD", passwd)
+
+    # These two urls are equivalent since we can pull the env vars out.
+    x = env_var_normalize("http://$USER:$PASSWORD@host/prefix/t/$TOKEN/suffix")
+    y = env_var_normalize(f"http://{user}:{passwd}@host/prefix/t/{token}/suffix")
+
+    assert x.env_var_url == y.env_var_url
+
+    replaced = y.conda_token_replaced_url()
+    assert replaced == f"http://{user}:{passwd}@host/t/<TOKEN>/prefix/suffix"
```
maresb, line 43: @mariusvniekerk, I'm trying to understand this code, and this looks wrong to me. Is there some crazy reason why `/prefix/` comes after and not before `<TOKEN>` in this assert?

Issue #170, opened 2022-03-11
Inconsistent ordering of lock files the first time the lock file is created compared to when the lock file already exists


lesteve:
Here is a bash script that illustrates this:
```bash
env_file=/tmp/environment.yml
lock_file=/tmp/conda-lock.yml

cat << EOF > $env_file
channels:
  - conda-forge

dependencies:
  - scikit-learn
EOF

function conda_lock_multiple_times() {
    rm -rf $lock_file
    for i in $(seq 1 4); do
        conda-lock lock --mamba -p linux-64 -f $env_file --lockfile $lock_file > /dev/null 2>&1
        echo "lock file md5sum: $(md5sum $lock_file)"
    done
}

echo "first run"
conda_lock_multiple_times

echo "second run"
conda_lock_multiple_times

echo "third run"
conda_lock_multiple_times
```

Output from the script:
```
❯ bash /tmp/test.sh
first run
lock file md5sum: e0b9c5509fa6f8732f07b6ee083442ee  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
second run
lock file md5sum: 6cebd9e7f6abfcadf260df476f2bb225  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
third run
lock file md5sum: a3e0986da5b6f4afcc23399b4612b397  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
lock file md5sum: 5fe5c3b275225e02abd17fca8529c4c3  /tmp/conda-lock.yml
```

[...]

PR #184, opened 2022-04-01
sort packages prior to rendering

### LINKED ISSUE #183: explicit lockfiles no longer sorted
Running the same test as #170, but with `--kind explicit` results in a different hash on every run:

```bash
env_file=/tmp/environment.yml
lock_file=/tmp/conda-explicit.lock

cat << EOF > $env_file
channels:
  - conda-forge

dependencies:
  - scikit-learn
EOF

function conda_lock_multiple_times() {
    rm -rf $lock_file
    for i in $(seq 1 4); do
        conda-lock lock -k explicit --mamba -p linux-64 -f $env_file --filename-template $lock_file > /dev/null 2>&1
        echo "lock file md5sum: $(md5 $lock_file)"
    done
}

echo "first run"
conda_lock_multiple_times

echo "second run"
conda_lock_multiple_times
```

 `0.13.2` produces consistent output everytime, while `1.0.1` and `1.0.4` produce different hashes on every run. I presume the fix is similar to #181, but on a different code path.

minrk:
ensures consistent package ordering, however lockfile object was produced

I'm guessing this is a redundant sort at least sometimes, but sorting immediately before generating output seems like the right place to ensure it always happens, unless topo sort is important earlier in the chain.

same issue as #170 in a different situation. fixes #183

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -555,6 +555,10 @@ def render_lockfile_for_platform(  # noqa: C901
 
     conda_deps = []
     pip_deps = []
+
+    # ensure consistent ordering of generated file
+    lockfile.toposort_inplace()
+
     for p in lockfile.package:
         if p.platform == platform and ((not p.optional) or (p.category in categories)):
             if p.manager == "pip":


mariusvniekerk:
It might be a little redundant, but the toposort is fast enough that running it multiple times shouldn't be problematic.

minrk:
#185 fixes the pre-commit bug

PR #186, opened 2022-04-17
Force resolving dependency via pip

olivier-lacroix:
Use the `[tool.conda-lock.dependencies]` section in the `pyproject.toml` file to force resolving a package via pip, fixing (part of) https://github.com/conda-incubator/conda-lock/issues/115.

This is useful in particular for packages that are not present on conda channels.

Note that

- I have used the `{source = 'pypi'}` syntax to flag a package which should be installed via pip to match how conda-locks interprets poetry sections. In case of poetry, `source` designate the pypi repository where the package should be downloaded. i.e. it could be a different (e.g. private) repository, in which case conda-lock should most likely also flag it as a pip installable-package. This is a question broader than this PR though, so I assume this should be OK here, and dealt with more broadly.
- If a package marked with `{source = 'pypi'}` in the `[tool.conda-lock.dependencies]` is not present in the packages dependencies (specified outside of `[tool.conda-lock.dependencies]`), it will not be added as a dependency. This makes sense to me as something downloaded via pip should be a python dependency. Currently it will be ignored silently - maybe a warning would be handy here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,13 @@ python = "3.9"
 ampel-ztf = {version = "^0.8.0-alpha.2", source = "pypi"}
 ```
 
+A dependency will also be treated as a `pip` dependency if explicitly marked with `source = "pypi"` in the `[tool.conda-lock.dependencies]` section, e.g.:
+
+```toml
+[tool.conda-lock.dependencies]
+ampel-ztf = {source = "pypi"}
+```
+
 In both these cases, the dependencies of `pip`-installable packages will also be
 installed with `pip`, unless they were already requested by a `conda`
 dependency.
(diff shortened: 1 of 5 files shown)


mariusvniekerk:
Can you add a test for this in the test suite?

olivier-lacroix:
@mariusvniekerk sure. Will do.

olivier-lacroix:
@mariusvniekerk apologies for not following-up sooner...

I have modified `test_parse_flit` to check that the manager is properly set to `pip` for relevant packages. Let me know if you think this is appropriate / enough in terms of tests

mariusvniekerk (approved):

PR #195, opened 2022-05-19
Fix deprecated keys in setup.cfg

ocefpaf:
~~Don't merge this yet. I'm trying to make it possible to install from the GitHub main branch too but I need to make a few extra adjustments.~~

Actually this is not needed, one can just install from "git" instead of a zipped git URL, from the error message:

```
For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
```

This is ready for review now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,7 @@
 from setuptools import setup
 
 
-setup(use_scm_version=True)
+setup(
+    name="conda_lock",
+    use_scm_version=True,
+)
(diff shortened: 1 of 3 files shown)


jezdez (approved):

setup.py
```diff
@@ -1,4 +1,7 @@
 from setuptools import setup
 
 
-setup(use_scm_version=True)
+setup(
+    name="conda_lock",
```
ocefpaf, line 5: Note that this is not really needed for the package but for GitHub. It cannot compute the projects that depends on this without the name in setup.py.

PR #201, opened 2022-06-25
Adding private pip repo support for legacy pypi repos.

shopigarner:
This PR adds private repo support for legacy pypi repositories allowing pip dependencies to be installed from private repos. This should enable most self-hosted private repos to be used (like Nexus and Artifactory). Please see the README for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,13 @@ The default category is `main`.
 [environment.yml][envyaml], using [Poetry's][poetry] dependency solver, if
 installed with the `pip_support` extra.
 
+### private pip repositories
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
Can you add an example of what the `pyproject.toml` should look like?

shopigarner:
@mariusvniekerk I've not tested it through a pyproject.toml as that's not my use case. This will only work with the global poetry config. If you have some suggestions on how to get it working with a `pypoetry.toml` file as well I'd be happy to try them out.

Issue #175, opened 2022-03-15
Error in case conda-lock is started from another drive than the default temporary directory lies on


wagnerpeer:
Hi there

I am using `conda-lock` for a while now and I am experiencing issues since the update to the new major version.

Issue
-----
If I run `conda-lock` from a path where the drive is different to the drive from the path of the lockfile I want to generate, the process crashes with a `ValueError: Paths don't have the same drive`.

Environment
--------------
Current `conda-lock`  version installed: `1.0.2=pyhd8ed1ab_0`
System used: `Windows 10 Enterprise 20H2`

Details
-------
I am using a full build-recipe for `conda-build`, a `meta.yaml` file as source.

The invocation is equivalent to:
```conda lock -p win-64 --file E:\recipe\meta.yaml --lockfile C:\Temp\lock.lock```

The issue lies in conda_lock\common.py, line 122, in the function `relative_path`. There is an implicit assumption, that the source and target paths have a common drive.

All `1.x.x` versions show this behavior, whereas the last prior version `0.13.2` is working as expected.

wagnerpeer:
Sure.

Full output from the mentioned command:

```shell
C:\Users\Username\Miniconda3\lib\site-packages\setuptools\command\install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
Traceback (most recent call last):
  File "C:\Users\Username\Miniconda3\lib\site-packages\pkg_resources\_vendor\packaging\requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "C:\Users\Username\Miniconda3\lib\site-packages\pkg_resources\_vendor\pyparsing.py", line 1654, in parseString
    raise exc
  File "C:\Users\Username\Miniconda3\lib\site-packages\pkg_resources\_vendor\pyparsing.py", line 1644, in parseString
    loc, tokens = self._parse( instring, 0 )
  File "C:\Users\Username\Miniconda3\lib\site-packages\pkg_resources\_vendor\pyparsing.py", line 1402, in _parseNoCache
    loc,tokens = self.parseImpl( instring, preloc, doActions )
  File "C:\Users\Username\Miniconda3\lib\site-packages\pkg_resources\_vendor\pyparsing.py", line 3417, in parseImpl
    loc, exprtokens = e._parse( instring, loc, doActions )
[...]

PR #204, opened 2022-06-30
Adding options to log addition additional metadata into the lockfile

NatPRoach:
Hello! thank you for making conda-lock, it's a great tool.

This PR would add the ability to specify extra comment lines into the lockfile as well as info about git repository etc. I added these metadata lines as a series of new flags:
```
    --metadata-jsons
    --metadata-yamls
    --metadata {"timestamp" "git_sha" "git_user_name" "git_user_email" "input_md5" "input_sha"}
```
`--metadata` can be specified multiple times to request any of the individual metadata tags, as requested here:
https://github.com/conda-incubator/conda-lock/pull/204#discussion_r996374759


mariusvniekerk:
Yeah having the metadata exist as a nice structured pydantic model with the various fields being Optional sounds great to me.  When converting the model to raw python dictionary you can easily just exclude the fields with `None` values

mariusvniekerk:
Can you add some tests to assert that the kind of metadata output we would want remains in the files.  These kinds of regression tests are invaluable to ensure that conda-lock has minimal regressions over time.

NatPRoach:
Absolutely. Just wanted to make sure that the approach looked reasonable to you before moving forward on writing tests. I'll try to get that done ASAP.

mariusvniekerk:
I landed #226 which should hopefully make the test suite here a bit more reliable

mariusvniekerk:
@wolfv Would this change impact micromamba in any way?

wolfv:
Let's ask @Klaim who implemented the lock-file handling in micromamba!

conda_lock/src_parser/__init__.py
```diff
@@ -101,6 +105,66 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
...
+class TimeMeta(StrictModel):
+    """Stores information about when the lockfile was generated."""
+
+    created_at: str = Field(..., description="Time stamp of lock-file creation time")
+
+    @classmethod
+    def create(cls) -> "TimeMeta":
+        import time
+
+        return cls(created_at=f"{time.asctime(time.gmtime(time.time()))}")
```
jezdez, line 117: Could you use an ISO 8601 formatted value here instead of `asctime` which is awkward to parse programmatically?

conda_lock/src_parser/__init__.py
```diff
@@ -101,6 +105,66 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
...
+            git_user_name=git.Git()().config("user.name"),
+            git_user_email=git.Git()().config("user.email"),
+            git_sha=git_sha,
+        )
+
+
+class InputMeta(StrictModel):
+    """Stores information about an input provided to generate the lockfile."""
+
+    md5: str = Field(..., description="md5 checksum for an input file")
```
jezdez, line 154: This seems useful, but shouldn't we also have a SHA256 value here?

jezdez (commented):
Just some nits with the formats and using md5 alone for the inputs metadata.

Klaim:
Hello!

> @wolfv Would this change impact micromamba in any way?

New fields in "metadata" will be ignored by micromamba (assuming I didn't miss a bug), see [details of the parsing in that function](https://github.com/mamba-org/mamba/blob/master/libmamba/src/core/env_lockfile.cpp#L78-L123). It will look for specific fields it needs.

tests/test-input-metadata/conda-lock.lock
```diff
@@ -0,0 +1,192 @@
...
+  - url: conda-forge
+    used_env_vars: []
+  - url: defaults
+    used_env_vars: []
+  content_hash:
+    linux-64: 1173e3c96ce20d063a5701b325b76deb97394f891af270af4ee0cb7cc1f6e838
+    osx-64: d01c1f5433f30bdbcd3bdad8d9b096774ab55f1210c094acdc61a35b32b28d67
+    win-64: 310b23581083bfb983927c40d3bdc86162192d7b26ffd7bffc385f627c155697
+  inputs_metadata:
+    /Users/c13371nroach/Documents/repos/conda-lock/tests/zlib/environment.yml:
```
mariusvniekerk, line 24: This should probably be a relative path not an absolute one

conda_lock/src_parser/__init__.py
```diff
@@ -126,6 +207,10 @@ def __or__(self, other: "LockMeta") -> "LockMeta":
             channels=self.channels,
             platforms=sorted(set(self.platforms).union(other.platforms)),
             sources=ordered_union([self.sources, other.sources]),
+            time_metadata=other.time_metadata,
+            git_metadata=other.git_metadata,
+            inputs_metadata=other.inputs_metadata,
```
mariusvniekerk, line 212: This should combine the inputs_metadata dictionaries as they probably point to different files

conda_lock/src_parser/__init__.py
```diff
@@ -126,6 +207,10 @@ def __or__(self, other: "LockMeta") -> "LockMeta":
...
+            custom_metadata=other.custom_metadata,
```
mariusvniekerk, line 213: This should also combine

conda_lock/conda_lock.py
```diff
@@ -745,13 +813,32 @@ def create_lockfile_from_spec(
         for dep in deps:
             locked[(dep.manager, dep.name, dep.platform)] = dep
 
+    git_metadata = GitMeta.create() if add_git_metadata else None
+    time_metadata = TimeMeta.create() if add_time_metadata else None
+    inputs_metadata: Optional[Dict[str, InputMeta]] = (
+        {str(src_file): InputMeta.create(src_file=src_file) for src_file in src_files}
```
mariusvniekerk, line 819: This should store the name as a relative path (maybe relative to the git root?) not absolute one

mariusvniekerk (changes requested):

conda_lock/src_parser/__init__.py
mariusvniekerk, line 117:
```suggestion
        return cls(created_at=datetime.datetime.utcnow().isoformat() + 'Z')
```

tests/test-input-metadata/conda-lock.lock
NatPRoach: I think (and I will x2 check this) That it uses a relative path if provided a relative path when conda-lock is invoked, and an absolute one if passed an absolute path. Happy to make the path relative (perhaps to the lockfile?) if you think that's always preferable.

conda_lock/src_parser/__init__.py
```diff
@@ -101,6 +105,66 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
...
+class TimeMeta(StrictModel):
+    """Stores information about when the lockfile was generated."""
+
+    created_at: str = Field(..., description="Time stamp of lock-file creation time")
+
+    @classmethod
+    def create(cls) -> "TimeMeta":
+        import time
+
+        return cls(created_at=datetime.datetime.utcnow().isoformat() + 'Z')
```
jezdez, line 117:
```suggestion
        return cls(created_at=datetime.datetime.utcnow().isoformat(timespec="seconds") + 'Z')
```

By default Python's datetime includes microseconds and that may not be useful to have a "portable" ISO format, plus the microseconds probably don't matter that much for our usecase. The [`timespec` parameter](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) was added in Python 3.6 to handle cases like that.

tests/test-input-metadata/conda-lock.lock
mariusvniekerk: I think that is generally preferable.  Leaking in usernames etc into the lockfiles seems undesirable

NatPRoach:
@mariusvniekerk @jezdez I believe this should be ready for re-review. I think I've addressed your comments. Apologies for delay.

conda_lock/conda_lock.py
```diff
@@ -745,13 +813,37 @@ def create_lockfile_from_spec(
...
+    inputs_metadata: Optional[Dict[str, InputMeta]] = (
+        {
+            relative_path(lockfile_path.parent, src_file): InputMeta.create(
+                src_file=src_file
+            )
+            for src_file in src_files
+        }
+        if src_files is not None
+        else None
+    )
```
riccardoporreca, line 827: I am happy to provide some input here after having fiddled around with consistency of source files in my recent PRs #230 #232

In general, we should make sure the inputs metadata are added to the lock file consistently with the existing `sources` key that conda-lock is already including in the metadata, which are passed as `sources` to `LockMeta` below

- As you can see, there is no need to have the `src_files` as an argument, since the same is available as `spec.sources`
- It is important to make sure paths are resolved in the same way in the `sources` and `inputs` metadata, which is not the case: `source.resolve()` below vs. `relative_path(lockfile_path.parent, src_file)

Something along the lines of the following might do the job

```py
def create_lockfile_from_spec(
    # [...]
    # boolean add_inputs_metadata replacing src_files
):
# [...]
    # assign resolved sources first (still with resolve(), this will be changed in #230 )
    spec_sources = {str(source.resolve()): source for source in spec.sources}
    inputs_metadata: Optional[Dict[str, InputMeta]] = (
        {
            resolved: InputMeta.create(src_file=src_file)
            for src_key, src_file in spec_sources.items()
        }
        if add_inputs_metadata
        else None
    )
    # [...]
    return Lockfile(
        package=[locked[k] for k in locked],
        metadata=LockMeta(
            # [...]
            sources=list(spec_sources.keys()),
            # [...]
```

NatPRoach: Hi Riccardo, previously @mariusvniekerk had indicated relative paths were preferable to him in the context of the lock-file. I'm not sure which person I should be listening to in order to get this merged, but it seems to me that `resolve` does the opposite of what @mariusvniekerk previously requested.  https://github.com/conda-incubator/conda-lock/pull/204#discussion_r933222792

riccardoporreca: Indeed, I agree that relative paths should be the way to go, which is also the reason for me opening #229 after the absolute paths were introduced in #189, probably too "wildly" to support #175

If @mariusvniekerk agrees, feel free to use relative paths in the aligned `spec_sources` hinted above, I will then rework my open PR #230, to re-handle different drives.

NatPRoach: It seems like #175 becomes a problem on windows tests with my implementation. Do you have a timeframe you hope to re-address #175 in #229?

riccardoporreca: Indeed, because the change for #175 was backed by unit tests that now fail. I would be ready anytime to change my proposed change for #229 in #230 to deal with the change you are introducing here (there is clearly a conflict between your branch and my branch).

My idea was to wait for this PR to be merged first and re-work #330 (aka rebase and solve conflicts) shortly after, but of course it is not great for your PR now with the failing test, unless @mariusvniekerk is fine to merge this despite the (motivated) failure.

One simple alternative could be the following: I propose my changes as PR to your branch in your repo, and @mariusvniekerk would integrate the fix for #229 together with your larger changes in the PR.

What do you think?

riccardoporreca: @NatPRoach, I was probably over-thinking... the easy way to work this out is to restore passing
`sources=[str(source.resolve()) for source in spec.sources]`

```py
    # [...]
    return Lockfile(
        package=[locked[k] for k in locked],
        metadata=LockMeta(
            # [...]
            sources=[str(source.resolve()) for source in spec.sources],
            # [...]
```
This is also pretty fair since what you are proposing in this PR was never meant to change any of the pre-exiting behavior. I will re-align the two as part of my PR.

riccardoporreca: @NatPRoach, sorry, I guess you need to take a further step back, since
```py
    spec_sources = {
        relative_path(lockfile_path.parent, source): source for source in spec.sources
    }
```
would still always be executed but does not handle yet the different drives (which again will be supported again with my PR).
We should scrap the `spec_sources` above an go for (note also the modified condition, fixing the wrong `add_inputs_metadata is not None`):
```py
    inputs_metadata: Optional[Dict[str, InputMeta]] = (
        {
            relative_path(lockfile_path.parent, src_file): InputMeta.create(
                src_file=src_file
            )
            for src_file in spec.sources
        }
        if add_inputs_metadata
        else None
    )
```
This is basically your very first code, just using `spec.sources` and replacing the argument with `add_inputs_metadata`

NatPRoach:
Hi @mariusvniekerk , @jezdez, et al. Just pinging again to remind that this could use another review whenever you get the chance.

NatPRoach:
Hi @mariusvniekerk  @jezdez , just wanted to ping about this one more time as I believe it's ready for re-review when you get the chance.

tests/test_conda_lock.py
```diff
@@ -481,6 +516,111 @@ def test_run_lock(
...
+    if is_micromamba(conda_exe):
+        monkeypatch.setenv("CONDA_FLAGS", "-v")
+    run_lock([zlib_environment], conda_exe=conda_exe, add_time_metadata=True)
+    end_time = datetime.datetime.utcnow()
+    lockfile = parse_conda_lock_file(TIME_DIR / DEFAULT_LOCKFILE_NAME)
+
+    time_metadata = lockfile.metadata.time_metadata
+    assert time_metadata is not None, "Time metadata was None"
+    assert (
+        start_time
```
mariusvniekerk, line 557: This could be <= as well right?  If you get lucky at the exact ms

conda_lock/src_parser/__init__.py
```diff
@@ -101,6 +120,89 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
...
+    )
+    git_user_email: Optional[str] = Field(
+        default=None, description="Git user.email field of global config"
+    )
+    git_sha: Optional[str] = Field(
+        default=None, description="sha256 hash of the most recent git commit"
+    )
+
+    @classmethod
+    def create(cls) -> "GitMeta":
```
mariusvniekerk, line 154: Do we want to have the git metadata associated with the HEAD commit for the relevant repo or should it be associated with the most recent commit for the input files.  

The latter would mean that if the components of a lock don't change then the git_meta is unaffected

maresb (commented):
Ideally we'd update the following commands in the comments at the top of the file so that they reproduce the correct metadata:

```
#     conda-lock lock --lockfile conda-lock.yml --update PACKAGE
#     conda-lock -f /home/mares/repos/conda-lock/tests/test-input-hash-zlib/environment.yml --lockfile conda-lock.yml
```

conda_lock/src_parser/__init__.py
```diff
@@ -101,6 +120,89 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
...
+            git_user_name = git.Git()().config("user.name")
+            git_user_email = git.Git()().config("user.email")
+        except git.exc.GitCommandError:
+            git_user_name = None
+            git_user_email = None
+        return cls(
+            git_user_name=git_user_name,
+            git_user_email=git_user_email,
+            git_sha=git_sha,
+        )
```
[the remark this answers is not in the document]
mariusvniekerk: Probably having the metadata selector fields behave similar to how channels work.  That jives with how the rest of the ecosystem functions.  So something like

```
cona-lock --md created-by --md comment --md command .... 
```

Fits most cleanly and works easily with click/typer which is what we're using for the cli

NatPRoach:
@mariusvniekerk  @maresb 
I believe I've addressed the remaining comments you had. Please let me know if there's anything else you'd like to see to get this merged.

mariusvniekerk:
Thanks!  Let me review this quickly

mariusvniekerk:
We probably don't need `--metadata-json` and `--metadata-yaml` as json is a subset of yaml

conda_lock/src_parser/__init__.py
```diff
@@ -113,6 +269,23 @@ class LockMeta(StrictModel):
...
+        default=None,
+        description=(
+            "Metadata dealing with the git repo the lockfile was created in and the user that created it"
+        ),
+    )
+    inputs_metadata: Optional[Dict[str, InputMeta]] = Field(
+        default=None,
+        description="Metadata dealing with the input files used to create the lockfile",
+    )
+    custom_metadata: Optional[Dict[str, str]] = Field(
```
maresb, line 285: Suppose I want to include something like `environment.yml` which contains lists. Shall we make this a fully general dict?

(Needs `from typing import Any`)

```suggestion
    custom_metadata: Optional[Dict[Any, Any]] = Field(
```

conda_lock/conda_lock.py
```diff
@@ -716,13 +732,43 @@ def _solve_for_arch(
...
+                f"Custom metadata field {key} provided twice, overwriting value "
+                + f"{to_change[key]} with {change_source[key]}"
+            )
+    to_change.update(change_source)
+
+
+def get_custom_metadata(
+    metadata_yamls: Sequence[pathlib.Path],
+) -> Optional[Dict[str, str]]:
+    custom_metadata_dict: Dict[str, Any] = {}
```
maresb, line 754:
```suggestion
) -> Optional[Dict[Any, Any]]:
    custom_metadata_dict: Dict[Any, Any] = {}
```

conda_lock/src_parser/__init__.py
mariusvniekerk: Ideally not.  It places a lot of extra strain on other implemations that make use of the file format like `micromamba`.  `Any` is FAR too broad a type.  I can potentially see value in a `Union[str, list[str]]`

conda_lock/conda_lock.py
mariusvniekerk: -1 for the same reason.  It basically means that this can no longer be serialized reliably etc

conda_lock/src_parser/__init__.py
maresb: Good point. Ideally it would be some sort of recursive type like `T: Union[int, str, list[T], dict[str,T]]`, but Pydantic chokes on that already.

Note how `requirements` must be at least `list[Union[str, dict[str, list[str]]]]` due to `["pip", {"pip": ["pypi-dep"]}]`. Thus to parse most `requirements.txt` we'd need at least something like
```
dict[str, Union[str, list[Union[str, dict[str, list[str]]]]]]
```
(assuming I didn't make any mistakes).

Given the complexity, I've convinced myself that this should be a separate feature. But in this case, we should probably make it clear in the `--help` that metadata-jsons/metadata-yamls contents must be `dict[str,str]`.

conda_lock/conda_lock.py
maresb: Updated for consistency:

```suggestion
) -> Optional[Dict[str, str]]:
    custom_metadata_dict: Dict[str, str] = {}
```
We may want to add runtime validation as well, since the validation error from Pydantic is really confusing.

conda_lock/src_parser/__init__.py
maresb: Do we want to support `dict[str, Union[str, int, float]]`?

mariusvniekerk:
probably sufficient as the first draft.  I want to make some tweaks to the models a little bit in a subsequent pr. And we can get the cli cleaned up in another one.

mariusvniekerk (approved):

PR #209, opened 2022-07-09
Bump CI from 3.8 -> 3.10

ericdill:

PR #220, opened 2022-07-13
Upgrade conda vendor

ericdill:
This moves from copy-pasted code from the conda git repo to a proper submodule. The reason for doing this is two-fold:

1. we now know which commit the conda code was copied from
2. we can now make use of git rebase commands to preserve the local changes necessary to vendor conda when upgrading the conda code itself. 

Given the failures in #209 , the current thinking is that the failures might just be resolved if we upgrade the vendored conda source code. So this is a useful first step in upgrading to the latest conda tag.

Note that this PR does not yet actually upgrade to the latest tag, it aims to be a like-for-like replacement to use submodules while maintaining the exact functionality. If this works then we can upgrade to the latest tag in a different PR


ericdill:
@mariusvniekerk feels like this test error is unrelated to the changes in this PR

```
@pytest.mark.timeout(90)
deftest_run_lock_with_update(monkeypatch, update_environment, conda_exe):
        monkeypatch.chdir(update_environment.parent)
if is_micromamba(conda_exe):
            monkeypatch.setenv("CONDA_FLAGS", "-v")
        pre_lock = {
            p.name: p
for p in parse_conda_lock_file(
                update_environment.parent / DEFAULT_LOCKFILE_NAME
            ).package
        }
        run_lock([update_environment], conda_exe=conda_exe, update=["pydantic"])
        post_lock = {
            p.name: p
for p in parse_conda_lock_file(
                update_environment.parent / DEFAULT_LOCKFILE_NAME
            ).package
        }
>       assert post_lock["pydantic"].version == "1.8.2"
E       AssertionError: assert '1.7' == '1.8.2'
E         - 1.8.2
E         + 1.7
/Users/runner/work/_temp/tests/test_conda_lock.py:480: AssertionError
```

maresb:
I think this has been superseded by #256

PR #224, opened 2022-07-20
Add support for ${ENV_VAR} syntax

jonashaag:


jonashaag:
@mariusvniekerk something seems off with CI that's unrelated to my changes.

mariusvniekerk:
@jonashaag Something doesn't seem quite right for the pattern matching.  For token matching it might be simpler to just include `{` and `}` to the matched character class than the lookahead thing you're trying here.

jonashaag:
I was missing a `?` after the `}` but I agree that we don't have to be so strict.

PR #226, opened 2022-07-28
Improved tests using the "test-update" directory

mariusvniekerk:
This should reduce some of the spurious errors observed in some test runs.

* Additionally reworked how test directories were made.  They should now all be fully isolated from each other
by making use of the tmp_path fixture.
* Added a futility used during the few install tests to ensure that we never run >1 install at a time as that
inevitably leads to download locking issues in conda/mamba


ericdill (approved):
first commit seems primarily adding types, formatting and removing lockfile usage. remaining commits seem straightforward enough. if there's anything hairy in that first commit that I should take a closer look at lmk, otherwise :shipit: imo

PR #230, opened 2022-08-10
Consistent lockfile-relative vs absolute sources

riccardoporreca:
Resolve #229

PR #232, opened 2022-08-10
Fix handling of default sources 

### LINKED ISSUE #231: Fix logic for running with out --file sources
Running w/o sources for dependencies is allowed by design (and would use the sources from the lockfile).

However, the CLI interface checks the existence of an `environment.y(a)ml` file for such cases, as this is the default:

https://github.com/conda-incubator/conda-lock/blob/7d9bd6d67d59fdd30d92a2ace3ee344aafa6a2b1/conda_lock/conda_lock.py#L1134-L1143

riccardoporreca:
Resolves #231, also
* making sure to handle separately the case of unspecified `--file` arguments from the case when `--file environment.y(a)ml` is explicitly used
* reviewing the checks for unspecified sources, also covering the usage of the defaults as fall-back when the lockfile exists

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -939,9 +939,18 @@ def run_lock(
                     file=sys.stderr,
                 )
         else:
-            long_ext_file = pathlib.Path("environment.yaml")
-            if long_ext_file.exists() and not environment_files[0].exists():
-                environment_files = [long_ext_file]
+            # bail out if we do not encounter any default .y(a)ml files
+            candidates = [
+                p if p.exists() else p.with_suffix(".yaml") for p in DEFAULT_FILES
+            ]
+            environment_files = [p for p in candidates if p.exists()]
+            if len(environment_files) == 0:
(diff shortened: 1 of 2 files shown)


riccardoporreca:
The failure on Windows seems to be for a "flaky" test, but I cannot re-run the job on this repo

riccardoporreca:
On hold waiting for PR #204, see https://github.com/conda-incubator/conda-lock/pull/204#discussion_r943911165

riccardoporreca:
@maresb, similar to #230, I would be happy to rebase this PR as #231 is still an issue with the current code-base.

Let me know if I can go ahead with this or I should rather wait for some ongoing work-in-progress to be first merged to the main branch.

maresb:
Sorry, no time to look at this in any detail this week, but generally speaking, if the issue isn't resolved, I'd recommend just going for the fix. We'll have to deal with any merge conflicts as they arise, likely via the greedy algorithm. Thanks a lot for all the work!!!

Issue #235, opened 2022-09-03
Straighten out Poetry dependency chaos


maresb:
I wonder if we need to depend on Poetry as a required dependency, since it shouldn't be necessary unless there's a `pip: [...]` section, right?

Poetry finally released a "stable" v1.2 a few days ago, but unfortunately not all is well. There are two blockers which currently make a conda-forge release impossible: 1. it contains a circular dependency with `poetry-plugin-export` which makes conda balk, and 2. there is no stable release of its dependency cleo v1.0.

In terms of testing, it's possible to install Poetry v1.2 / Cleo from conda-forge via dev labels.

In the meantime I've gotten fed up with Poetry, and discovered Hatch, which I like much better. (But it seems we are only using Poetry for its PyPI dependency solver, which Hatch doesn't have.) Currently I can't simultaneously install Poetry 1.1 and Hatch since they have [incompatible pins](https://github.com/pypa/hatch/issues/449) for the `packaging` dependency, and this means that I cannot simultaneously use recent versions of conda-lock with Hatch. :disappointed: 

So I have two questions:

1. Can we turn Poetry into an optional dependency?
2. How much must be changed here in conda-lock in order to accommodate Poetry v1.2?

maresb:
I was thinking that a good solution may be to cut out and vendor just the solver portion of Poetry which is needed by conda-lock. That would hopefully make things much more lightweight.

mariusvniekerk:
The other approach would be to make another pip resolver using the one from pip-tools (invocation here https://github.com/jazzband/pip-tools/blob/master/piptools/scripts/compile.py#L488-L499)

maresb:
Thanks for the tip!

I have not yet in the slightest understood these abstractions. I'll have a look and try to understand what's involved.

mariusvniekerk:
> * Can we turn Poetry into an optional dependency?
It already IS an optional dependency

mariusvniekerk:
If you want to rework the setup stuff to use `hatch` + `build` that would be fantastic

maresb:
Yes, great idea! How do you feel about tiangolo's Typer for CLI? (It's like Click the style of FastAPI with runtime type checking and Pydantic.)

mariusvniekerk:
Given that we already depend on both click and pydantic, all for that

maresb:
I feel like I successfully set up a default subcommand in Typer sometime in the past. But I'll verify that before I sink too much time. Thanks!

[...]

PR #239, opened 2022-09-15
Switch from setuptools to Hatch

maresb:


maresb:
I think we need to solve #235 in order to get the tests to pass.

But does this otherwise look good? There's obviously more to improve, but it seems to me like that can be done in separate PRs.

pyproject.toml
```diff
@@ -1,3 +1,88 @@
...
+    "Makefile",
+    "docs",
+    "docs/*",
+    "*.enc",
+    "tests",
+    "tests/*",
+]
+
+[tool.coverage.run]
+omit = "conda_lock/vendor/*"
```
mariusvniekerk, line 60: seems these should be a lists?

maresb: Ah yes, I meant to change it, but I was directly translating at first... commit pending...

maresb: Done

pyproject.toml
```diff
@@ -1,3 +1,90 @@
...
+
+[project]
+name = "conda_lock"
+description = "Lockfiles for conda"
+authors = [
+    { "name" = "Marius van Niekerk", email = "marius.v.niekerk@gmail.com" },
+]
+readme = "README.md"
+license = "MIT"
+requires-python = ">=3.6"
```
ofek, line 13: fyi Hatchling is 3.7+ so builder job can't use that

pyproject.toml
```diff
@@ -1,3 +1,90 @@
...
+
+[project.scripts]
+conda-lock = "conda_lock:main"
+
+[project.urls]
+Homepage = "https://github.com/conda-incubator/conda-lock"
+Repository = "https://github.com/conda-incubator/conda-lock"
+Documentation = "https://conda-incubator.github.io/conda-lock/"
+"Issue Tracker" = "https://github.com/conda-incubator/conda-lock/issues"
+PyPI = "https://pypi.org/project/conda-lock/"
```
ofek, line 41: nit: users will only ever be on that page when they see these

```suggestion
```

pyproject.toml
mariusvniekerk: but hatch can still build for older pythons right?  its just metadata

.github/workflows/test.yml
```diff
@@ -71,7 +70,6 @@ jobs:
           echo "${PATH}"
           which pip
           which python
-          python -m pip install -r requirements.txt
```
mariusvniekerk, line 74: pretty sure we need this for the poetry test dep?

pyproject.toml
ofek: yes

maresb: Ok, so we might need to adjust the workflows so that we build with 3.7+ before testing with 3.6.

But @mariusvniekerk, is there any benefit to supporting 3.6? Conda-forge stopped 3.6 builds some time ago. Plus, one could always `condax install` conda-lock to have it run in an isolated environment with modern Python.

pyproject.toml
maresb: Good point, thanks!

.github/workflows/test.yml
maresb: I removed it since `requirements-dev.txt` contains the line

```
-r requirements.txt
```

But perhaps we should instead remove that line and explicitly install those separately.

Or even better, we may want to leverage Hatch environments. But I'd prefer to do that in a separate PR. I'm trying to limit the current scope to switching from setuptools.

maresb:
Rebased on #241

mariusvniekerk:
Looks good in general, probably needs a rebase to get rid of the already added fixes for virtual packages

maresb:
All green now. (Seems there is a sporadic crash on the Windows job.)

mariusvniekerk:
Saw a sporadic crash on the osx on too on dev.  Mostly nothing to be too worried about

PR #240, opened 2022-09-17
Vendor Poetry

### LINKED ISSUE #235: Straighten out Poetry dependency chaos

maresb:
Closes #235

Still in early stages, but want to test with CI.

I haven't forgotten the [alternative proposal](https://github.com/conda-incubator/conda-lock/issues/235#issuecomment-1241985227) to write my own solver, but vendoring seems like the most straightforward place to start.


maresb:
Current goals:

* Get tests to pass
* Fix pre-commit
* Eliminate extraneous dependencies from vendored Poetry

maresb:
Rebased on #241 and cleaned up commits

maresb:
Weird, the `.whl` is missing a bunch of files which are included in the sdist... I'll have to look into this once I get the chance.

maresb:
I think the missing files may have simply been setuptools weirdness. Rebasing on Hatch, it looks like everything's there now.

maresb:
BTW, my vendored poetry is also subvendoring all sorts of stuff. :roll_eyes: Do I need to write a license aggregator now? :joy:

mariusvniekerk:
@maresb you can potentially make use of `vendoring` which is the thing that pip uses.  We can't use if for conda since conda can't easily be installed using pip, but for poetry it should work fine.  `vendoring` takes care of license aggregation etc.

maresb:
Unfortunately your advice came slightly too late. I'm just putting the finishing touches on a tool I made yesterday.

maresb:
This seems to be fundamentally working. But I still want to see if I can simplify things by using `vendoring`...

maresb:
Ok, this is almost ready to go. `vendoring` allowed me to make a few steps declarative rather than imperative, but it made a big mess of the subvendored licenses. I think I just need to do some quick refactoring into a cleanup script.

mariusvniekerk:
@maresb I've got a similar one in #256 that uses vendoring to update conda

mariusvniekerk (approved):

PR #249, opened 2022-09-30
add --version flag

### LINKED ISSUE #228: conda-lock --version doesn't work
I'm used to being able to check version numbers with

```bash
conda-lock --version
```

but unfortunately I get

```
Error: No such option: --version
```

bollwyvl:
## References

- fixes #228

## Changes
- [x] adds `click.version_option`
- [x] sort, normalize, and deduplicate `requirements.txt` 
- [x] test

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,11 @@
-PyYaml >= 5.1
-requests >=2
-Jinja2
-toml
-ensureconda >=1.3
 click >=8.0
 click-default-group
-pydantic >=1.8.1
+ensureconda >=1.3
+jinja2
 poetry <1.2
+pydantic >=1.8.1
+pyyaml >= 5.1
+requests >=2
 ruamel.yaml
-poetry
+toml
 typing-extensions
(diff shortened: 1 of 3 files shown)


bollwyvl:
Doesn't seem like the windows fail is related.

requirements.txt
```diff
@@ -1,12 +1,11 @@
...
-pydantic >=1.8.1
+ensureconda >=1.3
+jinja2
 poetry <1.2
+pydantic >=1.8.1
+pyyaml >= 5.1
+requests >=2
 ruamel.yaml
-poetry
+toml
```
bollwyvl, line 10: should probably start using `tomli`...

mariusvniekerk: What's the difference?

bollwyvl: sorry, confused with `pytoml`, which is definitely deprecated.

https://github.com/uiri/toml hasn't had a substantive PR merged in some time, and many open issues, but is probably still fine. 

I guess the biggest thing in favor of `tomli` is it's just the parsing side, and is the basis for the upcoming `tomllib` in stdlib: https://peps.python.org/pep-0680/#rationale

mariusvniekerk: If you want to make a or to shift to tomli (or I guess preferentially the stdlib one for py 3.11/12) that would be awesome.

PR #256, opened 2022-10-06
Make use of pip's tool vendoring to vendor conda

mariusvniekerk:


jezdez:
Nice!! <3

maresb:
@mariusvniekerk, why was Conda vendored from GitHub and not PyPI?

maresb:
Ah, because it's yanked from PyPI due to

> WARNING: Using `pip install conda` or `easy_install conda` will not give you conda as a standalone application. Currently supported install methods include the Anaconda installer and the miniconda installer. You can download the miniconda installer from https://conda.io/miniconda.html.

PR #258, opened 2022-10-09
Fix aggregating lock specs

### LINKED ISSUE #257: Dependencies missing when using multiple `--platform` flags with recipe
When locking a recipe (i.e. `meta.yaml`) with multiple `-p` flags, the locked dependencies are only valid for one of the platforms.

E.g. `conda-lock -p linux-64 -p osx-64` will only lock dependencies for either `linux-64` or `osx-64` but not both.

Example using the test `meta.yaml`
```
❯ conda lock -p linux-64 -p osx-64 -f tests/test-recipe/meta.yaml
...
Locking dependencies for ['linux-64', 'osx-64']...
INFO:conda_lock.conda_solver:linux-64 using specs []
INFO:conda_lock.conda_solver:osx-64 using specs ['python', 'pip', 'cython >=0.28.2', 'numpy', 'python-dateutil >=2.6.1', 'pytz >=2017.2', 'zlib', 'pytest']
...
```

As you can see in the example above, `linux-64` is not locked correctly.

I think the issue lies in incorrectly aggregating dependencies, will try to open a mini PR to fix this

bstadlbauer:
This should fix #257 where dependencies would only ever be locked for one platform for a recipe (`meta.yaml`), even when using multiple `-p` flags. 

Seems like this was due to incorrectly aggregating selectors when overriding an existing spec.

Fixes #257

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 
 ## Developing
 
-For the most up-to-date instructions see the github actions [test.yml workflow](./github/workflows.test.yml)
+For the most up-to-date instructions see the github actions [test.yml workflow](./.github/workflows/test.yml)
 
 1. Ensure conda and mamba are installed. Install [mambaforge](https://github.com/conda-forge/miniforge#mambaforge) if you're otherwise not sure which one to pick.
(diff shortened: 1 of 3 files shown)


bstadlbauer:
@mariusvniekerk Thanks for the quick merge and release!

PR #267, opened 2022-10-27
Support `conda/mamba install --copy`

mariusvniekerk:


docs/docker.md
```diff
@@ -15,24 +16,74 @@ You want a dockerfile that is structured something similar to this
...
+FROM condaforge/mambaforge:4.14.0-0 as builder
+
+COPY environment.yml /docker/environment.yml
+
+RUN . /opt/conda/etc/profile.d/conda.sh && \
+    mamba create --name lock && \
+    conda activate lock && \
+    mamba env list && \
+    mamba install --yes pip conda-lock>=1.2.2 setuptools wheel && \
+    conda-lock lock \
```
maresb, line 67: If I'm following correctly, then `conda-lock.yml` only lives in the Docker cache where it can't be tracked with version control. Do we really want to recommend this?

docs/docker.md
```diff
@@ -15,24 +16,74 @@ You want a dockerfile that is structured something similar to this
...
+# build the image
+docker build -t myimagename:mytag .
 ```
 
 This will ensure that your conda dependencies used in this docker container are
 always exactly reproducible.
+
+## conda-lock inside a build container
+
+You can also use conda-lock with a build-container style system if you mnake use of
```
maresb, line 51:
```suggestion
You can also use conda-lock with a build-container style system if you make use of
```

docs/docker.md
```diff
@@ -15,24 +16,74 @@ You want a dockerfile that is structured something similar to this
...
 # Dockerfile
 
-# Build container
-FROM continuumio/miniconda:latest as conda
+# -----------------
+# Builder container
+# -----------------
+FROM continuumio/miniconda:latest as builder
 
 ADD conda-linux-64.lock /locks/conda-linux-64.lock
```
maresb, line 24: I believe best practice is to prefer `COPY` over `ADD` when not using `ADD` features.

```suggestion
COPY conda-linux-64.lock /locks/conda-linux-64.lock
```

docs/docker.md
```diff
@@ -15,24 +16,74 @@ You want a dockerfile that is structured something similar to this
...
+
+RUN . /opt/conda/etc/profile.d/conda.sh && \
+    conda activate lock && \
+    conda-lock install \
+        --mamba \
+        --copy \
+        --prefix /opt/env \
+        /docker/conda-lock.yml
+# optionally you can perfom some more cleanup on your conda install after this
+# to get a leaner conda environment
```
maresb, line 81: Let's just be explicit here?

```suggestion

RUN mamba clean --all --yes
```

conda_lock/conda_lock.py
```diff
@@ -1206,6 +1215,11 @@ def lock(
     default=False,
     help="don't attempt to use or install micromamba.",
 )
+@click.option(
+    "--copy",
+    is_flag=True,
+    help="Install using `--copy`.  This is useful for building containers",
```
maresb, line 1221: Just because I didn't know offhand what `--copy` does.

```suggestion
    help=(
        "Install using `--copy` to prevent links. "
        "This is useful for building containers"
    ),
```

docs/docker.md
mariusvniekerk: hence the two approaches in this

maresb: I feel like I'm missing something obvious here... I think I don't know what is a "build-container style system".

Two reasons I can think of off the top of my head for locking:
1. Control explicitly when dependencies are updated
2. Having an explicit record of the versions installed

It seems like neither of these applies in this case. (For 2. at least there's a file in the container, but if you have to go into the container to see it, you might as well run `conda env export`.)

What's the benefit to this approach over installing directly from `environment.yml`?

mariusvniekerk: So the main benefit is more when you are using something like `earthly` and have a single earthfile specify multiple docker outputs.  

1. This example does NOT deal with this case
2. The explicit record is at least kept in the build-container style build described here, which means you can introspect a bit more easily as to what should be in the env.

maresb:
BTW, I recently implemented another variation [here](https://github.com/conda-forge/automerge-action/pull/16/files), actually inspired by this PR, where I use micromamba in a build Docker-stage to generate the Conda environment and then copy the Conda environment into the primary Docker-stage. (As a result, I reduced the image size of the CF automerge bot. :smile:)

mariusvniekerk:
> BTW, I recently implemented another variation [here](https://github.com/conda-forge/automerge-action/pull/16/files), actually inspired by this PR, where I use micromamba in a build Docker-stage to generate the Conda environment and then copy the Conda environment into the primary Docker-stage. (As a result, I reduced the image size of the CF automerge bot. 😄)

This would probably make for a nicer example

maresb:
> This would probably make for a nicer example

Cool, I'll aim to add a MWE along these lines this weekend.

Issue #264, opened 2022-10-26
Resolving pip dependencies with uppercase characters fails


HedgehogCode:
If I have pip dependencies that have uppercase characters in the name `conda-lock` will fail with a `KeyError`. Note, that pip is case insensitive. Therefore it is easy to fix the environment definition by not using upper-case letters. However, `conda-lock` should also be case insensitive when handling pip dependencies. 

Try the following environment definition (`env.yml`):
```yaml
name: myenv
channels:
- conda-forge
dependencies:
- pip
- pip:
  - MIDIFile
```

Error:
```
$ conda-lock lock -f env.yml
Locking dependencies for ['linux-64', 'osx-64', 'win-64']...
INFO:conda_lock.conda_solver:linux-64 using specs ['pip *']
Traceback (most recent call last):
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/conda_lock/src_parser/__init__.py", line 283, in seperator_munge_get
    return d[key]
KeyError: 'MIDIFile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/conda_lock/src_parser/__init__.py", line 286, in seperator_munge_get
    return d[key.replace("-", "_")]
KeyError: 'MIDIFile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/bin/conda-lock", line 10, in <module>
    sys.exit(main())
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/benjamin/miniconda3/envs/conda-lock-dev/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #268, opened 2022-10-28
lowercase pip package names

mariusvniekerk:
This should hopefully address #264

### DIFF
--- a/conda_lock/lookup.py
+++ b/conda_lock/lookup.py
@@ -23,11 +23,13 @@ class _LookupLoader:
         res = requests.get(self._mapping_url)
         res.raise_for_status()
         self._pypi_lookup = yaml.safe_load(res.content)
-        # lowercase the pypi names
+        # lowercase and kebabcase the pypi names
         assert self._pypi_lookup is not None
-        self._pypi_lookup = {k.lower(): v for k, v in self._pypi_lookup.items()}
+        self._pypi_lookup = {
+            k.lower().replace("_", "-"): v for k, v in self._pypi_lookup.items()
+        }
         for v in self._pypi_lookup.values():
-            v["pypi_name"] = v["pypi_name"].lower()
+            v["pypi_name"] = v["pypi_name"].lower().replace("_", "-")
 
     @property
     def pypi_lookup(self) -> Dict[str, MappingEntry]:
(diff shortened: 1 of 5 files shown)


maresb:
`environment.yaml` is missing

tests/test_conda_lock.py
```diff
@@ -184,6 +184,12 @@ def channel_inversion(tmp_path: Path):
     )
 
 
+@pytest.fixture
+def env_with_uppercase_pip(tmp_path: Path):
+    """Path to an environment.yaml that has a hardcoded channel in one of the dependencies"""
+    return clone_test_dir("test-uppercase-pip", tmp_path).joinpath("environment.yaml")
```
maresb, line 190:
```suggestion
    return clone_test_dir("test-uppercase-pip", tmp_path).joinpath("environment.yml")
```

conda_lock/lookup.py
```diff
@@ -3,27 +3,42 @@
...
 
     def resolve(self) -> None:
         res = requests.get(self._mapping_url)
         res.raise_for_status()
         self._pypi_lookup = yaml.safe_load(res.content)
+        # lowercase the pypi names
+        assert self._pypi_lookup is not None
+        self._pypi_lookup = {k.lower(): v for k, v in self._pypi_lookup.items()}
+        for v in self._pypi_lookup.values():
+            v["pypi_name"] = v["pypi_name"].lower()
```
bollwyvl, line 30: perhaps also normalize `_` to `-`?

maresb:
It should work with py311, no? I think Mamba was migrated yesterday.

mariusvniekerk:
> It should work with py311, no? I think Mamba was migrated yesterday.

Given the conda-forge a bit more time with the great rebuild before i move the ci to also try 3.11.  It almost certainly works, just not sure if all the dependencies we need exist there yet

conda_lock/lookup.py
mariusvniekerk: yeah that feels prudent

bollwyvl: Here's "the law": https://peps.python.org/pep-0503/#normalized-names

```
import re

def normalize(name):
    return re.sub(r"[-_.]+", "-", name).lower()
```

PR #271, opened 2022-10-31
Fix `--auth-file` flag

### LINKED ISSUE #270: `--auth-file` contents are not used for `conda-lock install`
There's a typo in the `install` command, we should check for the combined `_auth` object instead of looking only at the value of the `auth` flag:

https://github.com/conda-incubator/conda-lock/blob/04e24386914158e10f6c60084655fdf207df008d/conda_lock/conda_lock.py#L1268-L1270

https://github.com/conda-incubator/conda-lock/blob/04e24386914158e10f6c60084655fdf207df008d/conda_lock/conda_lock.py#L1284-L1286

simonbohnen:
Closes #270

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -1281,7 +1281,7 @@ def install(
     with _render_lockfile_for_install(
         lock_file, include_dev_dependencies=dev, extras=set(extras)
     ) as lockfile:
-        if auth:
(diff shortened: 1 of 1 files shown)


maresb:
Just a personal style nit, we could make this `if _auth is not None:` to be more explicit, right?

maresb (approved):

PR #276, opened 2022-11-08
Recommend installing with pipx or condax

maresb:
As per https://github.com/conda-incubator/conda-lock/issues/275#issuecomment-1307230432

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,14 @@ various places.
 
 ## Installation
 
-Use one of
+Use *one* of the following commands:
 
 ```bash
 pipx install conda-lock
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,11 +22,17 @@ various places.
 
 ## Installation
 
+Use one of
+
 ```bash
+pipx install conda-lock
+condax install conda-lock
 pip install conda-lock
 conda install -c conda-forge conda-lock
```
mariusvniekerk, line 31: Should we also recommend 

```suggestion
conda install -c conda-forge -n base conda-lock
```

PR #280, opened 2022-11-14
Stress that .lock extension is necessary

klieret:
It took me a while to realize that conda/mamba refuses to deal with conda-lock files if they have a yaml ending (throwing a very ugly decoding error).

This PR tries to make this more clear in the readme (though it is more of an issue with conda/mamba, I guess).

(Also see https://github.com/mamba-org/provision-with-micromamba/issues/106)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,8 @@ conda-lock --lockfile superspecial.conda-lock.yml
 ```
 
 The extension `.conda-lock.yml` will be added if not present. Rendered
-environment files (env or explicit) will be named as as
-`"conda-{platform}.lock"`.
+environment files (env or explicit) must end with `.lock` and will be named as
+`"conda-{platform}.lock"` by default.
 
 If you want to override that call conda-lock as follows.
 ```bash


maresb (approved):
Thanks a lot for this! I think such small documentation changes make a big difference.

mariusvniekerk:
We can also potentially just attempt to parse every file initially with the conda lock format and if that works treat it as such. That minimizes surpise factors

klieret:
My major dissatisfaction was that the error message was very unhelpful. So just catching & reraising the error with a bit of added bit of information ("Please note that file extensions matter: For example, if you try to read a .... file... use the conda-lock.yml ending") etc. would already add a lot, without changing any logic.

maresb:
@klieret, thanks again for the contribution. We're very happy to accept PRs which improve the error handling. For example, I just added #282 this morning.

klieret:
Thanks @maresb!  I plan  follow up with a PR (though I'm a bit short on time atm)

PR #288, opened 2022-11-24
Do not use `--kind=explicit`

sebastian-luna-valero:
Fixes https://github.com/conda-incubator/conda-lock/issues/285

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -21,6 +21,10 @@ dependencies:
 If in this case `some_pip_only_library` depends on `requests` that dependency will be met by
 conda and the version will be constrained to what the conda solver determines.
 
+Please note that explicit lock files do not support installing `pip` dependencies.
+The unified lockfile format does (i.e. the `--kind=explicit` flag cannot be used
+with `conda-lock --file=environment.yml`)
+
 ## Usage with pyproject.toml
 
 If a dependency refers directly to a URL rather than a package name and version,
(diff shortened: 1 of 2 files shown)


docs/pip.md
```diff
@@ -21,6 +21,10 @@ dependencies:
 If in this case `some_pip_only_library` depends on `requests` that dependency will be met by
 conda and the version will be constrained to what the conda solver determines.
 
+Please note that explicit lock files do not support installing `pip` dependencies.
+The unified lockfile format does (i.e. the `--kind=explicit` flag cannot be used
+with `conda-lock --file=environment.yml`)
```
maresb, line 26: To address https://github.com/conda-incubator/conda-lock/issues/285#issuecomment-1326553508, 

```suggestion
We recommend avoiding the `--kind=explicit` flag when there are `pip`
dependencies. Most tools (except for `conda-lock install`) do not recognize
`pip` dependencies from explicit lockfiles, so they may be silently ignored.
The default lockfile format explicitly supports pip dependencies.
```

maresb (changes requested):

maresb (approved):

maresb:
We could also omit a warning from `conda-lock` itself. How about
```python
        if len(pip_deps) > 0:
            logger.warning(
                "WARNING: installation of pip dependencies is only supported "
                "by the 'conda-lock install' command. Other tools may silently "
                "ignore them. For portability, we recommend using the newer "
                "unified lockfile format (i.e. removing the --kind=explicit "
                "argument."
            )
```

right [here](https://github.com/conda-incubator/conda-lock/blob/e1feddd058940dca4227e00927ff61e84901a2eb/conda_lock/conda_lock.py#L677)?

sebastian-luna-valero:
Good idea, done!

PR #289, opened 2022-12-01
Add "platform_system" to the environment when solving in Pypi

romain-intel:
Without it, the following YAML:
```
channels:
  - conda-forge
dependencies:
  - pip
  - pip:
    - twisted
```
Will result in bringing in twisted-iocpsupport when solving on Linux despite the fact that that package is for Windows only.

This was due to the platform_system marker missing.

This PR also adds linux_{arch} for Linux platforms which is similar to what packaging does (see https://github.com/pypa/packaging/blob/main/packaging/tags.py#L434).


romain-intel:
Thanks for merging!

PR #297, opened 2022-12-14
make gitpython dependency optional

### LINKED ISSUE #296: Make gitpython dependency optional?
#204's adding the hard dep on `gitpython` lights up vulnerability scanners for `CVE-2022-24439`. Here's the upstream issue:

- https://github.com/gitpython-developers/GitPython/issues/1515

As there's no particular timeline for a fix, perhaps the `gitpython` dependency could be made optional, as with `pip` for the non-vendored bits of `poetry`?

bollwyvl:
## References
- fixes #296

## Changes
- move `gitpython` from `requirements.txt` to `requirements-dev.txt`
- return `None` when creating `GitMeta` if `gitpython` is not installed

### DIFF
--- a/conda_lock/src_parser/__init__.py
+++ b/conda_lock/src_parser/__init__.py
@@ -169,7 +169,10 @@ class GitMeta(StrictModel):
         metadata_choices: AbstractSet[MetadataOption],
         src_files: List[pathlib.Path],
     ) -> "GitMeta | None":
-        import git
+        try:
+            import git
+        except ImportError:
+            return None
 
         git_sha: "str | None" = None
         git_user_name: "str | None" = None
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -840,11 +840,6 @@ def create_lockfile_from_spec(
     else:
         git_metadata = None
 
-    git_metadata = GitMeta.create(
-        metadata_choices=metadata_choices,
-        src_files=spec.sources,
-    )
-
     if metadata_choices & {MetadataOption.InputSha, MetadataOption.InputMd5}:
         inputs_metadata: Optional[Dict[str, InputMeta]] = {
             relative_path: InputMeta.create(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dynamic = ["dependencies", "version"]
 license-files = { paths = ["LICENSE"] }
 
 [project.optional-dependencies]
+git_support = ["gitpython"]
 pip_support = ["poetry ==1.1.*"]
 
 [project.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ dynamic = ["dependencies", "version"]
 license-files = { paths = ["LICENSE"] }
 
 [project.optional-dependencies]
-git_support = ["gitpython"]
 pip_support = ["poetry ==1.1.*"]
 
 [project.scripts]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -27,3 +27,4 @@ mkdocs-click
 mkdocs-include-markdown-plugin
 flaky
 docker
+gitpython
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,6 @@ pyyaml >= 5.1
 ruamel.yaml
 tomli; python_version<"3.11"
 typing-extensions
-gitpython
 toolz >=0.12.0,<1.0.0
 # poetry:
 cachecontrol[filecache] >=0.12.9


conda_lock/src_parser/__init__.py
```diff
@@ -169,7 +169,10 @@ def create(
         metadata_choices: AbstractSet[MetadataOption],
         src_files: List[pathlib.Path],
     ) -> "GitMeta | None":
-        import git
+        try:
+            import git
+        except ImportError:
+            return None
```
maresb, line 175:
```suggestion
            raise RuntimeError("The GitPython package is required to read Git metadata.")
```

bollwyvl: Wait, that's probably not right, as `GitMeta.create` gets [created unconditionally](https://github.com/conda-incubator/conda-lock/blob/90b370a7b8f2b918820159a3c4f95ac0a31ac446/conda_lock/conda_lock.py#LL826C5-L826C17).

maresb: Right.. it's not computed on-demand like I originally thought, so this is going to require more changes. :slightly_frowning_face:

bollwyvl: rolled back... would a `warning` be more appropriate?

maresb: I feel like this should be a fatal error; the metadata is explicitly requested, so I think it's much better to hard fail. Here's another attempt:
https://github.com/bollwyvl/conda-lock/pull/1

I still need to run pytest on it though.

maresb: Running tests via CI here:
https://github.com/conda-incubator/conda-lock/pull/298

bollwyvl: It's already marked optional all over: as in, `conda-lock` doesn't fail in a non-git checkout.

maresb: Right, but if a user runs
```
conda-lock --metadata git_user_name
```
then it would be mean to the user to silent fail, right? Or am I missing something obvious?

bollwyvl:
Sure, so maybe the fields need to be lazy, then, and explode if None
because of missing gitpython.

maresb:
> so maybe the fields need to be lazy

That's what I thought at first, but I believe https://github.com/bollwyvl/conda-lock/pull/1 suffices.

maresb:
Argh, sorry, not sure what was wrong with me last night... :dizzy_face: 

Could you please delete the redundant `get_metadata = ...` block immediately under what I added in the PR?

maresb:
Perfect, thanks!!!

maresb:
What are your thoughts regarding the future of the GitPython dependency? Namely, suppose someone fixes the vulnerability. Then are we okay to reinclude it as a required dependency? It seems to be fairly lightweight and pure-Python.

The reason I ask is that if we can reinclude it after some time, then perhaps we could avoid introducing `git_support`. (It will soon become yet another thing we need to deprecate.)

bollwyvl:
Since you asked: I'd prefer `conda-lock` had zero dependencies, which is of course unreasonable. I use it as the "bootstrap" layer for a lot of things, and every extra dep is another thing to track that could inject weirdness, get a CVE, etc. 

Specifically for git: if using `conda-lock`, I check in the locks, so having git metadata in them is... not very useful from an entropy point of view.

As for the metadata stuff in general: Any time i can avoid invoking _any_ conda/mamba stuff, I do. Basically 100% of the time, I strip out the conda-lock integrity hash, because it's basically impossible to reconstruct what it would be, and doesn't particularly help git diffs (ha!). If anything, I'll inject the sorted set of depdendencies as a comment, and then run/not-run conda-lock if that has changed.

As to the other optional: I don't use the `pip` stuff at all (pypi/packaging/poetry/etc. has been so flaky of late), or the new yaml format. It's just another thing that reduces the portability vs. systems that already have `conda` or whatever.

So sure, it's fine if a non-vulnerable `gitpython`,  comes back, and pip extras are of course hot trash, but I'd really rather only pay for what I need.

maresb:
Thanks a lot for typing that all out, I really appreciate the perspective.

>  I'd prefer conda-lock had zero dependencies, which is of course unreasonable.

:+1: My motivation for vendoring Poetry was to remove a dependency, and hopefully eventually eliminate it entirely.

> pypi/packaging/poetry/etc. has been so flaky of late

:+1: Though I don't have many complaints about Hatchling so far.

> or the new yaml format

Any way to win you over on this? It definitely requires some more work for stability, but Micromamba support is to me a killer feature.

Improving stability and reducing dependencies are high on my todos here, but those unfortunately require me to somehow find some weekend days... :laughing:

maresb:
For this PR, would you be amenable to dropping `git_support`?

bollwyvl:
> Any way to win you over on this?

Probably not? My use cases are highly automation/forensics driven, and _minimally entropic, reusable files on disk, checked into git_ are pretty much my happy place. Parsing YAYAML (Yet Another YAML) format without a JSON schema, vs _a list of URLs_ is pretty much no contest for me.

> Micromamba support is to me a killer feature

But `micromamba`, along with every version of `conda` back to the oldest running CI machine, _also_ supports `--kind=explicit`. As do other tools, such as [jake](https://github.com/sonatype-nexus-community/jake) (sorta).

> find some weekend days

don't i know it, friend! But conda-lock is a critical daily driver for me, so I'm happy to help when I can.

bollwyvl:
> Hatchling

I'm happy it's working for you!

And you _didn't_ ask, but: to me, the cambrian explosion of build systems vs the cartesian product of duplicative plugins is extremely tiresome. If important information doesn't appear _in pyproject.toml_ in the _standards compliant fields_, I feel like the point has partially been missed. The failure modes of `python -m build` (or whatever is the _non_-arbitrary-code-running way to invoke it), with the layers of tracebacks, are getting to nodejs-levels of ick. Oh, and then there's the plugins that _install_ node, or other language runtimes, inside some magical hidden place. That's the best.

maresb (approved):

maresb:
ping @mariusvniekerk

Issue #278, opened 2022-11-11
Lockfile implementation of categories instead of category


maresb:
In order to make it possible to, for instance, reliably install packages in both `dev` and `main`, I proposed in https://github.com/mamba-org/mamba/issues/1209 changing the `category` attribute from a string to a list of strings called `categories`.

There hasn't been any movement yet (which I'm aware of) towards an implementation. I have an idea which might offer an easy way forward, purely in conda-lock at first...

What would happen if we duplicate the entries in the lockfile, so that the lockfile looks roughly like:?

```yaml
- name: common-package
  category: dev
- name: common-package
  category: main
```

If we are lucky, then perhaps the conda-lock and micromamba implementations will follow the expected behavior. (To be tested!) This way we wouldn't need a new lockfile format version.

If this works, this is still rather redundant, but not catastrophically so. We should aim to do better in lockfile format v2. In the meantime, since the micromamba folks believe that extra attributes will be ignored, we could start generating interoperable lockfiles which look like

```yaml
- name: common-package
  category: dev
  categories:
  - main
  - dev
- name: common-package
  category: main
  categories:
  - main
  - dev
```

so that this can be read as v1 by ignoring `categories`, or alternatively read as v2 by ignoring `category` (effectively containing duplicate entries). This way, we could first develop everything in `conda-lock`, and then only once we're ready, work on implementing v2 in micromamba.

srilman:
Is there an existing branch or implementation to test?

maresb:
Sorry, not yet. I just did by hand the experiment with micromamba described above. Would you like to try an implementation?

srilman:
Sure! One question: how should duplicate entires in separate input files be handled. Say, for example, we are running `conda-lock -f main.yml -f test.yml` on the following instances:

```
# main.yml
category: main
dependencies:
  python 3.10
  pandas 1.4
```

```
# test.yml
category: test
dependencies:
  pandas 1.5
```

Currently, Conda Lock with solve for `['python 3.10.*', 'pandas 1.5']` and produce a lock-file where Pandas's category is `test`. Should it be in both `main` and `test` or just `test`? What if `pandas 1.5` was specified in `main.yml`; should it be in both categories then?

I would argue that it should be only in the `test` category in the first case, but both categories in the second case.

maresb:
As written, that would be a dependency conflict.

The idea is to merge the dependencies into a single simultaneous solve. So what would happen logically is:

```yaml
dependencies:
[...]

Issue #222, opened 2022-07-20
Document (revise?) how dependencies from file sources are combined


riccardoporreca:
When processing file sources, it looks like only the last specification of a conda package version (by package name) is preserved

e.g.
```yml
# environment-foo.yml
channels:
- conda-forge
dependencies:
- python=3.8.12
- numpy=1.22.3
```

```yml
# environment-bar.yml
channels:
- conda-forge
dependencies:
- python=3.8.12
- numpy=1.22.2
```

```sh
conda-lock -f environment-foo.yml -f environment-bar.yml -p linux-64
INFO:conda_lock.conda_solver:linux-64 using specs ['python 3.8.12.*', 'numpy 1.22.2.*']
```

```sh
conda-lock -f environment-bar.yml -f environment-foo.yml -p linux-64
INFO:conda_lock.conda_solver:linux-64 using specs ['python 3.8.12.*', 'numpy 1.22.3.*']
```

Is this by design? Would it be possible to resolve to the union of the specified dependencies?
The behavior, if unchanged, should be best documented in [Compound specification](https://github.com/conda-incubator/conda-lock/blob/42923b3bd280a260f1480d45a5c7104546e44580/README.md#compound-specification)

The relevant code seems to be

https://github.com/conda-incubator/conda-lock/blob/42923b3bd280a260f1480d45a5c7104546e44580/conda_lock/src_parser/__init__.py#L336-L346

(As a side note, it looks lines 341-342 are not effective given line 344)

riccardoporreca:
Thanks for clarifying this @mariusvniekerk, happy to draft a documentation PR if you agree.

PR #300, opened 2022-12-18
Merge Version Constraints from Multiple Input Files

srilman:
Adds version merging behavior discussed in #278. Currently, if multiple input files contain the same dependency with different version constraints,`conda-lock` will use the version constraints from the last file passed in. This PR has `conda-lock` combine / AND the constraints together.

Thus, if there are a pair of incompatible version constraints from 2 different files (say `>=1,<2` and `=2`), `conda-lock` will now fail.


srilman:
FYI I am still working on this, but I am running into issues running some of the unit tests. conda-lock seems to be hanging when solving for pip dependencies. This hasn't happened before, so I am a bit lost. Since last working on this PR, I did install poetry on my machine, but I can't imagine how that would cause conflicts.

Planning to use Docker to get a new dev environment set up and make some progress on this.

maresb:
The tests are unfortunately somewhat flaky. It doesn't necessarily mean that the problem is on your side.

I find the previous implementation of some stuff fairly suspect. For example,

```python
            previous_selectors |= dep.selectors
```

This looks like it actually modifies the `prev_dep` object. I can't imagine that this was intended...

Luckily, `|=` (implemented by `Selectors.__ior__`) seems to only occur in this one place. It seems to me like we may want to eliminate and reimplement it.

Also, I don't think it even makes sense to merge different selectors. For instance,

```yaml
- python =3.9 # [win]
- python =3.10 # [linux]
```

seems like a perfectly reasonable specification which can't be merged. (Or am I just confused?)

Also, we may want to raise an error on mixing URLDependencies and VersionDependencies, or when mixing two URLDependencies which are not equal.

maresb:
Those preexisting issues aside, it looks really excellent! Thanks so much for putting this together. Thanks for your patience in waiting for review. I hope we can get this in soon.

srilman:
> Also, we may want to raise an error on mixing URLDependencies and VersionDependencies, or when mixing two URLDependencies which are not equal.

Good point, will add.


> Also, I don't think it even makes sense to merge different selectors. For instance,
> ```
> - python =3.9 # [win]
> - python =3.10 # [linux]
> ```
> seems like a perfectly reasonable specification which can't be merged. (Or am I just confused?)

No, I think you're correct. They shouldn't be merged together. But what about selectors that partially overlap? For example,
```
python >=3.7
python <3.11  #[win]
```
should technically be transformed to something like
```
python >=3.7, <3.11 #[win]
python >=3.7  #[not win]
```

How do we want to handle this case?

srilman:
Also, the way `conda-lock` currently generates lock specs for environment files is by creating a spec per platform and merging them together. For each spec, the selector for every dependency is just the current platform being rendered. So we rely heavily on taking the union of dependencies specs with different selectors.

maresb:
> should technically be transformed to something like
> 
> ```
> python >=3.7,<3.11  #[win]
> python >=3.7        #[not win]
> ```

Excellent point, I agree. Although I'm not sure if we want to consider this as an explicit "transformation". Ideally, in my mind, the platforms should somehow be computed independently.

> How do we want to handle this case?

Great question. I think it requires much more careful consideration than I can afford tonight. It would be nice to avoid any convoluted constructions. Probably someone will soon want to extend the current functionality of selectors. (For instance, I don't think conda-lock supports `not` in selectors.) So ideally we shouldn't implement anything that will make that difficult.

srilman:
I rewrote this PR to work on top of #316 and to reuse some helper functions from the `conda` vendor package.
- Now, dependency versions will only be merged if they are multiple versions for the same platform. This works due to the new SourceFile approached implemented in the previous PR
- This uses the `treeify` and `untreeify` functions from the `conda` vendor package to combine conda version strings. This is a lot simpler than parsing code from the previous implementation.

riccardoporreca:
@maresb, @srilman, linking here a question on the current behavior: https://github.com/conda/conda-lock/issues/222#issuecomment-1190286711

> This is by design. Files specified later are intended to be overrides to the specifications earlier

The change proposed here can be potentially breaking, and I have been in fact leveraging this behavior for a project where I want to provide an additional file overriding specific package versions. Would you consider an opt-in possibility of re-enabling the current behavior?

maresb:
Thanks @riccardoporreca for bringing this to our attention.

My primary motivating use case is to be able to specify dev dependencies alongside project dependencies, and so I want them to combine rather than override.

I can see that overriding versions is a valid method of combining dependencies, but it seems less motivated to me. Can you explain a little bit more about why you want to override version numbers in a way that's incompatible with the previous spec?

In any case, opting in to avoid a breaking change seems quite sensible. @mariusvniekerk, does it make sense to you to support combining deps, or are we off-track?

riccardoporreca:
> I can see that overriding versions is a valid method of combining dependencies, but it seems less motivated to me. Can you explain a little bit more about why you want to override version numbers in a way that's incompatible with the previous spec?

I do agree that combining all dependencies w/o overrides seems more reasonable in general, which is also what made me a bit surprised in the first place and motivated me to ask #222. Then (perhaps my bad), I though I could leverage the current "by design" feature to dynamically create and use an additional `environment-overrides.yml` in a Continuous Integration pipeline, where we want to assess different version of certain dependencies besides those specified in _nominal_ environment YAML files (and already locked): In practice, we re-run `conda-lock` on-the-fly with an additional `environment-overrides.yml`, specified last.

I also understand this is a pretty specific use-case, for which we can also foresee an alternative approach at our end if an opt-in possibility is not viable / not worth maintaining.

knedlsepp:
Found this issue while trying to split the constraints from my `conda-env.yml` into a separate file and was suprised that `conda-lock` silently ignored the version constraints for some packages. It was somewhat surprising that this doesn't work. Apparently this comes from the fact that in https://github.com/conda/conda-lock/blob/c18d8f2a3c172584a1b8fee2233f767e158fb3f5/conda_lock/src_parser/aggregation.py#L37 we just overwrite dependencies we already have in the list even if that means overwriting the `version` field with an empty string.

@maresb Any way we can move this forward?

maresb:
@knedlsepp, I'd love to get some help with this. See also #390 and #410.

Some good news is that the codebase has improved in significant ways since the previous discussion. Especially relevant is that platforms are now computed independently, and that was a big blocker before.

I think this is fundamentally doable now. Would you be up for starting a new PR?

PR #303, opened 2022-12-31
Typing improvements

maresb:
I had to clarify some things about the various types in order to start reviewing #300. Along the way I documented my findings as type annotations.


srilman (approved):
This looks great! I was definitely a bit confused looking at the PyPi solver, so good to see this change.

mariusvniekerk (approved):

PR #304, opened 2023-01-04
Add pyproject.toml option to prevent requests on pypi.org

yjeand:
My corporate proxy does not allow access to `pypi.org`. We have a private mirror accessible, but unfortunately `conda-lock` attempts to make requests to `pypi.org` even when the private mirror is set up, resulting in a conda-lock exception.

With this CLI flag, we can ensure that no request goes to `pypi.org`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,8 @@ Right now `conda-lock` only supports [legacy](https://warehouse.pypa.io/api-refe
 poetry config repositories.foo https://username:password@foo.repo/simple/
 ```
 
+The private repo will be used in addition to `pypi.org`. For projects using `pyproject.toml`, it is possible to [disable `pypi.org` entirely](#disabling-pypiorg).
+
 ### --dev-dependencies/--no-dev-dependencies
 
 By default conda-lock will include dev dependencies in the specification of the lock (if the files that the lock
@@ -390,6 +392,14 @@ ampel-ztf = {source = "pypi"}
 In both these cases, the dependencies of `pip`-installable packages will also be
 installed with `pip`, unless they were already requested by a `conda`
 dependency.
+
+#### Disabling pypi.org
+
+When using private pip repos, it is possible to disable `pypi.org` entirely. This can be useful when using `conda-lock` behind a network proxy that does not allow access to `pypi.org`.
+```toml
+[tool.conda-lock]
+allow-pypi-requests = false
+```
 
 ## Dockerfile example
 
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -742,6 +742,7 @@ def _solve_for_arch(
             conda_locked={dep.name: dep for dep in conda_deps.values()},
             python_version=conda_deps["python"].version,
             platform=platform,
+            allow_pypi_requests=spec.allow_pypi_requests,
         )
     else:
         pip_deps = {}
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -178,6 +178,7 @@ def solve_pypi(
     conda_locked: Dict[str, lockfile.LockedDependency],
     python_version: str,
     platform: str,
+    allow_pypi_requests: bool = True,
     verbose: bool = False,
 ) -> Dict[str, lockfile.LockedDependency]:
     """
@@ -199,6 +200,8 @@ def solve_pypi(
         Version of Python in conda_locked
     platform :
         Target platform
+    allow_pypi_requests :
+        Add pypi.org to the list of repositories (pip packages only)
     verbose :
         Print chatter from solver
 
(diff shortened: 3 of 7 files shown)


yjeand:
Thank you very much for your answer!

I agree that the ideal behavior would be to prevent unecessary requests to pypi.org without having to use an additional option. At first glance, the fact that pypi is the last item in the repositories list could be enough... but this is not what I see in my use case.

Here is an excerpt from my logs, including the commandline used and the stacktrace.
[stacktrace.txt](https://github.com/conda-incubator/conda-lock/files/10359338/stacktrace.txt)

Thank you for your time, please let me know if I can help in any way!

maresb:
Thanks a lot @yjeand for the stack trace and your willingness to help!

We are currently using a [vendored](https://github.com/conda-incubator/conda-lock/tree/main/conda_lock/_vendor/poetry) copy of Poetry which is [v1.1.15](https://github.com/python-poetry/poetry/tree/1.1.15). In your PR here you have identified the block where we initialize the `Pool` object with

```python
pool = Pool(repositories=[*repos, pypi])
```

Based on your stack trace, it seems that the `Pool` object is searching all the provided repositories. On the other hand, it seems that there are [ways to disable PyPI from Poetry](https://python-poetry.org/docs/repositories/#disabling-the-pypi-repository). So I would ask: does your Poetry configuration allow you to prevent Poetry from accessing pypi.org? If so, can we simply make our current code respect this setting? For instance, I notice that poetry itself never specifies the `repositories` argument to the `Pool` constructor.

yjeand:
Thank you for this analysis. For my project we currently use Poetry for our linux builds (I'm working on conda-lock support for Windows, and maybe to replace Poetry for Linux down the road). I can confirm that on our Linux builds Poetry (used directly, outside conda-lock) does not attempt to access pypi since we have added a source with `default = true` in our `pyproject.toml` file. However, conda-lock does not seem to transmit this setting to the vendored version of Poetry.

I will take a look into it and report back here.

yjeand:
Hello,

Here is a quick summary of my experimentations. No ideal solution has come up.

I think that the most natural behaviour for Poetry-based projects would be for conda-lock to respect the sources declared in `pyproject.toml` for Poetry. If a source has `default = true` then conda-lock could deactivate pypi.org (or let the Poetry solver do it). However at the moment conda-lock does not use those sources from `pyproject.toml` at all, and I do not feel capable of making such a big change. Thinking about it a bit more, this approach would be probably awkward for `pyproject.toml`-based projects which use a build system other than Poetry.

Conda-lock instead relies on globally defined sources in the user configuration file for Poetry, [like explained here](https://github.com/conda-incubator/conda-lock#private-pip-repositories). And as far as I know, there is no way to flag a repo as `default` when using `poetry config` like this (I have tried manual modifications of the config file, it did not work). If we add a way to flag a repo as "default" in conda-lock, I feel like we are getting back to a solution roughly equivalent to the original `--no-pypi` flag of my pull request in terms of user impact.

Coming at it from another angle, I wonder if it would not be simpler to just test a pypi.org connection from conda-lock to decide if we add it to the repositiories list that is passed to the vendored Poetry solver. Something along the lines of:

```python
    try:
        PyPiRepository(disable_cache=True).get_package_info("pip")
        repos.append(PyPiRepository())
    except RequestException as e:
        print(f"Could not access pypi.org, error was:\n{e}", file=sys.stderr)
    pool = Pool(repositories=[*repos])
```
in `pypi_solver.solve_pypi()`
(I'm not sure if the cache is reset each time conda-lock is run, so I deactivated it just in case)

What do you think about all this?

maresb:
> Thinking about it a bit more, this approach would be probably awkward for pyproject.toml-based projects which use a build system other than Poetry.

This is an excellent insight.

Perhaps it would make more sense to store conda-lock configuration related to project.toml in a `[tool.conda-lock]` section inside pyproject.toml itself. This would help keep the CLI clean. But I don't think we want to use it as any sort of primary configuration source, so this should somehow be self-evident. Maybe we could do

```toml
[tool.conda-lock.pyproject]
use-pypi = true
```

How do you feel about this idea?

yjeand:
Good idea, it would certainly work for my project.
I will try to implement it, see how it goes.

maresb:
For a subsequent PR, it might be interesting to deprecate `--extras`, `--filter-extras`, and `--pypi_to_conda_lookup_file` in favor of a similar `pyproject.toml`-based config.

yjeand:
Here is my attempt at implementing the `pyproject.toml`-based config to disable `pypi.org`.
I have named the setting differently from your example because it seemed more natural to me, but feel free to change it.

maresb:
This looks excellent @yjeand, thanks so much!!!

On second thought, I agree with you that `[tool.conda-lock]` is the correct level for this setting.

One final request: could you please add a short note to the README as documentation? After that, this looks ready from my perspective.

@mariusvniekerk, could you please have a look?

yjeand:
I have added a couple notes in the README.

maresb (approved):

PR #310, opened 2023-01-11
Attempt to upgrade poetry version

mariusvniekerk:
Attempt to upgrade our vendored poetry to latest


mariusvniekerk:
@maresb tried to run the vendoring helper, but it seems that it didn't replace the import paths appropriately

maresb:
Hi, sorry for the confusion.

The original vendoring procedure was quite complicated. My migration script was designed to run once, and only with that specific version of Poetry/Poetry Core/Cleo. (The idea was to implement it in code so that when it comes time to upgrade we know exactly what procedure was used, and we can make the appropriate changes.) Thus the migration named "Vendor Poetry" has already been applied.

The `migrate-code reset` command runs `git reset` to undo previously-applied migration commits, but this only works when `HEAD` is a migration commit.

Rather than add an `EPOCH` I'd suggest `m = Migration("Upgrade poetry to v1.3.2")`.

Taking a quick look, seems that we need to change `vendor_helpers.py` line 104 to

```python
            f"/{self.name.replace('-', '_')}-{self.version}.tar.gz"
```

and line 390 to

```python
                {"python_version": python_version, "sys_platform": "linux"},
```

but then it's getting confused due to importlib_metadata having multiple `python_version` requirements. I'll try to look into this soon.

maresb:
Superseded by #637

Issue #299, opened 2022-12-15
Deprecate pip_support optional dependency group


maresb:
After vendoring poetry in #240, the `pip_support` [option](https://github.com/conda-incubator/conda-lock/blob/90b370a7b8f2b918820159a3c4f95ac0a31ac446/pyproject.toml#L31) adds a superfluous install of Poetry.

Probably we should change this to an empty list and add a comment that it's a legacy artifact. (I don't think we should remove it because that would break lots of stuff... I'm curious if anyone has any clever ideas for how to do a deprecation warning. The only idea off the top of my head would be to create a dummy package on PyPI, which seems like an ugly solution.)

maresb:
Closed in #313

PR #313, opened 2023-01-12
Add filelock as a dependency

mariusvniekerk:
This seems to be needed for some classes of pip based resolution

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "hatchling.build"
 name = "conda_lock"
 description = "Lockfiles for conda"
 authors = [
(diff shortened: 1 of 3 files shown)


maresb:
As of [four months ago](https://github.com/ionrock/cachecontrol/pull/284), `filelock >=3.8.0` should already be [implicit](https://github.com/ionrock/cachecontrol/blob/c05ef9eff1c9ac176481fb99e7a7188aa5b4e17b/setup.py#L25) in `cachecontrol[filecache]`. But explicit is better than implicit.

I'm concerned about removing the `pip-support` option and breaking pipelines, but a quick experiment shows that pip only emits a warning, so I think we're good.

maresb (approved):

maresb:
Closes #299

PR #316, opened 2023-01-16
New Source File and Lock Specification Approach

srilman:


mariusvniekerk:
Can you break this into a few smaller prs doing the model moves first.  The way this is now is very hard to review

srilman:
#320 contains some of the refactoring I did in this PR.  Its about 500 lines of changes, but most of it is moving classes and functions around with no content changes.

maresb:
Wonderful. Now we need a rebase here, whenever you get the chance, and then we can review.

conda_lock/conda_lock.py
```diff
@@ -243,44 +236,6 @@ def fn_to_dist_name(fn: str) -> str:
     return fn
 
 
-def make_lock_spec(
```
srilman, line 246: Both `make_lock_spec` and `parse_source_files` are specifically related to parsing source files, and don't have a big effect on the rest of the program. Thus, I moved them to `conda_lock/src_parser/__init__.py` to reduce the amount of code in this file (since its about 1500 lines of code).

srilman:
I rewrote #300 to work on top of this PR. So we would need this PR in first.

conda_lock/conda_lock.py
maresb: Thanks a lot for all this work!!! It would still really help for reviewing to have these refactor steps in separate commits so that I could view the substantive changes separately. (In general it's much easier to follow several smaller logical commits than one massive one.)

I'm sincerely very eager to see this through quickly, but my schedule looks difficult at the moment. I'll see what I can do, but apologies in advance if I'm slow to respond.

srilman: @maresb I tried to break this PR down into a couple of commits to make it a bit easier. I had some trouble breaking down the last couple of commits since the contents is very much tied together. But if it is still difficult to look through, let me know.

conda_lock/src_parser/pyproject_toml.py
```diff
@@ -213,14 +215,14 @@ def specification_with_dependencies(
 
     if force_pypi:
         for dep in dependencies:
-            if dep.name in force_pypi:
-                dep.manager = "pip"
+            if dep.dep.name in force_pypi:
+                dep.dep.manager = "pip"
```
maresb, line 219: Having `dep.dep` and `.to_source()` seems a bit awkward. I wonder if there's a more natural approach to SourceDependency?

conda_lock/src_parser/__init__.py
```diff
@@ -59,32 +73,66 @@ class URLDependency(_BaseDependency):
 Dependency = Union[VersionedDependency, URLDependency]
 
 
+class SourceDependency(StrictModel):
+    dep: Dependency
+    selectors: Selectors = Selectors()
```
maresb, line 78: Can you explain in words what's the idea behind a SourceDependency?

maresb (commented):
I've started review. This looks like some really great work. However, I still have a ways to go before I get a grasp on the main commit 0ad6f7b. For now I have included some initial comments.

conda_lock/conda_lock.py
srilman: I see, good point. I'm used to writing large PRs in general. In the future, I can definitely break down my commits even further. Would you like me to modify the last large commit of this PR? My concern with modifying that commit is that I'm not sure how to break it down without having some commit be broken. I normally try to ensure that every commit exposed to master is a somewhat-working impl of the app or library.

conda_lock/src_parser/__init__.py
```diff
@@ -136,3 +138,85 @@ def aggregate_lock_specs(
...
+    Parameters
+    ----------
+    src_files :
+        Files to parse for dependencies
+    platform_overrides :
+        Target platforms to render environment.yaml and meta.yaml files for
+    """
+    from conda_lock.src_parser.environment_yaml import parse_environment_file
+    from conda_lock.src_parser.meta_yaml import parse_meta_yaml_file
+    from conda_lock.src_parser.pyproject_toml import parse_pyproject_toml
```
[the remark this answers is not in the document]
srilman: Those modules import `conda_lock/src_parser/__init__.py`. When I included them at the top, I was getting a circular dependency error. Those files use the `SourceDependency`, `SourceFile`, `VersionedDependency`, and `URLDependency` classes. If I move these classes to a file like `conda_lock/src_parser/models.py`, then I can get rid of the circular dependency and have these be top level imports. Thoughts?

conda_lock/src_parser/pyproject_toml.py
```diff
@@ -380,3 +342,41 @@ def parse_pdm_pyproject_toml(
     res.dependencies.extend(dev_reqs)
 
     return res
+
+
+def parse_pyproject_toml(
```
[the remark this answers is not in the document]
srilman: Mainly for code-discoverability. The general pattern in most of the modules is that the last function in a module is the main function exposed by that module and used by the rest of the program. Since `parse_pyproject_toml` is the main exported function, I thought it should be at the end to fit that pattern.

conda_lock/src_parser/pyproject_toml.py
srilman: For the `dep.dep` case, we can use properties to expose common attributes of `Dependency` objects like `name`, `manager`, etc.

conda_lock/src_parser/__init__.py
srilman: Sure (I can also add a comment). So we want the LockSpecification to be a mapping from environment (platform right now, but potentially other attributes in the future) to a list of dependencies. Those dependencies don't need selectors, because selectors are only used when constructing the LockSpecification for determining if a dep is required for an environment.

Thus, a SourceDependency represents a dependency with any additional info associated with it from a source file. Right now, its just selectors, but in the future, we may have other limiters like min or max Python version.

conda_lock/conda_lock.py
maresb: Large PRs are fine, it's just large commits which are difficult to understand.

Your commits don't need to be perfect, and I'm asking for a fairly high standard. I can explain how I try to write commits:

> I'm not sure how to break it down without having some commit be broken. 

This is a good rule in general. But in some situations I think it's fine to break something in one commit and fix it in a subsequent one. (For example, in one commit I might remove functionality X, and then in the next commit I add functionality Y which replaces X. This way the new details of Y aren't confused with the old details of X.)

What also helps is to stage partial changes. For example, in the process of implementing X, I may modify some type hints in another part of the code. In this case, I can stage and commit the type hints in a separate commit, so that my implementation of X remains focused.

The most complicated technique is to rebase code you've already committed, but that is really a lot of work.

A few concrete suggestions for how you could break up the main commit:
* Switch to `ruamel`
* Define new classes
* Implement core logic using the new classes

I think I can handle the large commit as-is, but I would need to find an uninterrupted block of time where I could work through the whole thing at once. If you manage to break up the commit, then I can probably finish reviewing it sooner.

conda_lock/src_parser/pyproject_toml.py
maresb: Would it make sense to have `SourceDependency` as a subclass of `Dependency`?

conda_lock/src_parser/__init__.py
maresb: Would something like this make sense as a docstring?

> A `SourceDependency` represents information about a particular dependency specification which has been extracted from a `SourceFile` (e.g. `environment.yml` or `pyproject.toml`). Generally, information about the target environment (e.g. platform) is not included, but it might be specified by including selectors.

Please rewrite in case I misunderstood the details.

srilman: That's perfect! Will add

conda_lock/src_parser/pyproject_toml.py
srilman: We'd need a `SourceURLDependency` and a `SourceVersionedDependency` in that case, right?

conda_lock/conda_lock.py
srilman: Have you spent a lot of time looking at the last 2 commits: `Initial Version of SourceFile Approach` and `Adding Test Yaml Files`. If not, I can try to split those further.

conda_lock/src_parser/__init__.py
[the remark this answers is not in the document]
srilman: It's not just for type hints. I will move them to a new module.

conda_lock/conda_lock.py
srilman: I think you already looked through the first 2 commits thoroughly, so I will leave them alone.

maresb: > I think you already looked through the first 2 commits thoroughly, so I will leave them alone.

Yes, in fact, if you create a separate PR for those I think we can already merge them since they are minor refactoring changes.

For the big commit I was thinking: since this is a major change, Marius will have to review it after me. Thus it may be worthwhile to invest extra time to make it readable.

conda_lock/src_parser/__init__.py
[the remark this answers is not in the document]
maresb: I forgot to mention in my previous comment that Python does permit certain types of circular imports.

Running `from a import B` is essentially equivalent to `import c; B = c.B`. Python will allow you to import `c` from `a` while also importing `a` from `c`, as long as you don't access the module attributes (i.e. `c.B`) before the module has been fully loaded. (This works when all `a.B` accesses occur within function definitions.)

So there is another strategy: rather than fix the cycles, try to import modules lazily, i.e. replace `from c import B` → `import c` and `B` → `c.B`. But I have the impression that this leads to very fragile code. I think it's generally much more robust to remove all non-typing-related circular imports when possible.

PR #321, opened 2023-01-23
fix #292: by not exporting empty CONDA_OVERRIDE_* variables

### LINKED ISSUE #292: KeyError if using recent conda 22.11.1 release
I noticed conda-lock CI jobs failing recently, which appears to be due to the latest `conda=22.11` [release from last week](https://github.com/conda/conda/releases) since I've confirmed that conda `22.9.0` still works as expected.  

With a fresh conda-lock 1.2.1 environment (in which conda=22.11.1 gets installed) there is the following error:

>  "error": "KeyError(PackageRecord(_hash=-5824671606007306977, name='__glibc', version='0', build='0', build_number=0, channel=Channel(\"@\"), subdir='linux-64', fn='__glibc', md5='12345678901234567890123456789012', package_type='virtual_system'))",

You can reproduce this with the following:
```yml
# environment.yml
channels:
  - conda-forge
dependencies:
  - gdal
platforms:
  - linux-64
```

```bash
docker run -it --rm condaforge/mambaforge:latest /bin/bash
mamba create -n condalock conda-lock mamba
conda activate condalock
conda-lock lock -f environment.yml
```

croth1:
Should be one way to fix #292 - now conda=22.11 produces the same lockfile as conda=22.9.

### DIFF
--- a/conda_lock/invoke_conda.py
+++ b/conda_lock/invoke_conda.py
@@ -158,7 +158,11 @@ def _process_stdout(stdout: IO[str]) -> Iterator[str]:
 
 
 def conda_env_override(platform: str) -> Dict[str, str]:
-    env = dict(os.environ)
+    env = {
+        key: val
+        for key, val in os.environ.items()
+        if not (key.startswith("CONDA_OVERRIDE_") and val == "")
+    }
     env.update(
         {
             "CONDA_SUBDIR": platform,


croth1:
unfortunately, this is not the fix: 
> CONDA_OVERRIDE_GLIBC - GLIBC version number or set to "" for no GLIBC. This only applies on Linux.
https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages

PR #323, opened 2023-01-29
Apply auth also for private pypi packages

croth1:
previous logic converted all lock files to explicit type and ignored all lines starting with `#`or `@`.  This however does not allow adding auth to packages from private pypi indices, as they are represented as `# pip package @ index_url` and thus were skipped in the auth re-adding logic.


croth1:
@maresb whom do I need to bribe to get this reviewed? :)

maresb:
Hi @croth1! I don't have much time these days, but at least this one is nice and short!

You could start by bribing me with a test case so that we can prevent regressions. :slightly_smiling_face: Thanks!

croth1:
@maresb, I integrated the new logic into existing test cases - strictly speaking the auth stripping logic does not seem to be needed for pypi, as the current poetry code does that automatically.
However this way the pypi auth logic is the same as the conda auth logic and was a bit easier to integrate into the tests. Hope that's fine with you.

conda_lock/conda_lock.py
```diff
@@ -91,10 +91,14 @@
 DEFAULT_FILES = [pathlib.Path("environment.yml")]
 
 # Captures basic auth credentials, if they exists, in the second capture group.
```
maresb, line 93:
```suggestion
# Captures basic auth credentials, if they exists, in the third capture group.
```

conda_lock/conda_lock.py
```diff
@@ -91,10 +91,14 @@
...
-AUTH_PATTERN = re.compile(r"^(https?:\/\/)(.*:.*@)?(.*)")
+AUTH_PATTERN = re.compile(r"^(# pip .* @ )?(https?:\/\/)(.*:.*@)?(.*)")
+
+# Do not substitute in comments, but do substitute in pip installable packages
+# with the pattern: # pip package @ url.
+PKG_PATTERN = re.compile(r"(^[^#@].*|^# pip .*)")
 
 # Captures the domain in the second group.
```
maresb, line 100:
```suggestion
# Captures the domain in the third group.
```

maresb (changes requested):
Thanks so much @croth1 for your persistence and patience on this!!!

(I wasn't so familiar with this code, so even though this is straightforward, it's been tricky for me to find the time to look properly at this. Not to mention that the current directory structure for these test files confuses me; perhaps we should eventually clean that up.)

This looks really excellent. Could you please add test cases for `test__add_auth_to_line`, `test__strip_auth_from_line`, `test__extract_domain`? Luckily these tests are self-contained, so I think this will be quick and easy.

Once that's done, it looks good from my side. Anything to add, @mariusvniekerk?

maresb (approved):

maresb:
Thanks so much @croth1!!!

PR #327, opened 2023-02-03
Update+apply pre-commit, prepare for pre-commit.ci, add badge

dbast:
### Description

Update+apply pre-commit hooks (isort requires an update to work again).

This also prepares for the usage of pre-commit.ci:
* Usually faster than the pre-commit Github Action with Cache 
* Enabled also for several other repos in the conda organisation 
* Shifts some CI load off from Github Actions (CI in the conda org sometimes is clogged due many conda/conda-build PRs with their huge/long-running Testsuites)
* pre-commit.ci also has an autofix feature which is disable in the config instead commenting "pre-commit.ci autofix" in a PR is enough.

I am (as member of the @conda/infrastucture team) happy to enable pre-commit.ci for this repo, if the @conda/conda-lock agrees to the idea of using pre-commit.ci. That said, I am also fine with descoping the PR to updates only and sticking to GH Actions.


ocefpaf (approved):

maresb (approved):
Generally I prefer the transparency of GHA, but for the case of pre-commit, I agree that it probably makes more sense to use pre-commit.ci.

Also, we should probably merge this sooner rather than later to get the isort fix.

dbast:
Ok. Going to enable pre-commit.ci for the repo (and changing the required status checks) ... and close / re-open the PR should then show that it works.

dbast:
Had to fix the error:
```
tests/conftest.py: error: Source file found twice under different module names: "conftest" and "tests.conftest"
tests/conftest.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
tests/conftest.py: note: Common resolutions include: a) adding `__init__.py` somewhere, b) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
``` 

Added a `__init__.py` to the tests folder ... this seems to have triggered mypy to scan files it previously didn't scan, due to pre-commit being configured to only scan changed files... so I had to exclude  `tests/test_conda_lock.py` from mypy as currently 73 errors.

see also https://github.com/python/mypy/issues/4008

This is now green on pre-commit.ci and locally with `pre-commit run --all` (which wasn't the case before).

maresb:
Thanks a lot for this! All green now. I noticed one more local mypy error on my side, corrected in https://github.com/dbast/conda-lock/pull/2

maresb:
Now that CI has finished, I'll merge it in this branch...

maresb:
ping @mariusvniekerk

Issue #229, opened 2022-08-09
Allow relative paths for sources


riccardoporreca:
Since #189, paths stored in the lockfile are stored as absolute, to address the issue in #175 where the lockfile and the sources are not on the same drive.

However, there was probably a point in relative paths and the change has its drawback for other (more common?). Consider sources and lockfiles maintained under version control in the same repo: one wouldn't want to produce and push lockfiles containing references to the absolute path of the local checkout.

I think using absolute paths should be opt-in, or used in case the construction of relative paths does not work.

riccardoporreca:
A sensible approach might be to restore the old behavior unless the source path is absolute, so absolute source paths (possibly including the drive) would stay as absolute. This should still address #175 but keep the (more sensible?) behavior for the other cases

riccardoporreca:
Actually, #189 might have introduced an inconsistency with the logic (from https://github.com/conda-incubator/conda-lock/pull/124/commits/df566dcfdd1ba4e127efe27649748a088b573d97) for reconstructing the source file path from the lock file itself if not specified

https://github.com/conda-incubator/conda-lock/blob/7d9bd6d67d59fdd30d92a2ace3ee344aafa6a2b1/conda_lock/conda_lock.py#L920-L932

which is probably broken: I haven't tested it, but I assume "reconstruct native paths" would fail for absolute paths in `lock_content.metadata.sources`.

I think the "sensible approach" hinted above would be easily to mirror in a consistent way in this logic as well. I will try to draft a PR.

JessicaS11:
Just wanted to leave a note that I support an opt-in (or at the very least optional) use of absolute paths. In our use case, we have multiple contributors who might update the lock files locally, and with absolute paths `conda-lock.yml` lists each person's local path as source (rather than just the shared environment.yml). Thanks for working on an update for this!

maresb:
@JessicaS11 I'm curious about your use case for opt-in of absolute paths. Would you want the path to the source file `environment.yml` to be absolute, or the `conda-lock.yml` in the comment header, or both?

Issue #155, opened 2022-02-24
poetry SolverProblemError when trying to lock an environment containing a matplotlib pip dependency


lesteve:
```yml
# /tmp/environment.yml
channels:
  - conda-forge
dependencies:
  - python=3.9
  - pip
  - pip:
    - matplotlib
```

```
conda-lock lock -p linux-64 -f /tmp/environment.yml 
```

Error:
```
poetry.puzzle.exceptions.SolverProblemError: Because -dummy-package- depends on matplotlib-base (*) which doesn't match any versions, version solving failed.
```

Probably there is some conversion logic between conda-forge and pip package names that kicks in although it should not?

Full output:
```
❯ conda-lock lock -p linux-64 -f /tmp/environment.yml 
Locking dependencies for ['linux-64']...
Traceback (most recent call last):
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/poetry/puzzle/solver.py", line 233, in _solve
    result = resolve_version(
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/poetry/mixology/__init__.py", line 7, in resolve_version
    return solver.solve()
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/poetry/mixology/version_solver.py", line 83, in solve
    self._propagate(next)
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/poetry/mixology/version_solver.py", line 123, in _propagate
    root_cause = self._resolve_conflict(incompatibility)
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/poetry/mixology/version_solver.py", line 321, in _resolve_conflict
    raise SolveFailure(incompatibility)
poetry.mixology.failure.SolveFailure: Because -dummy-package- depends on matplotlib-base (*) which doesn't match any versions, version solving failed.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/local/lesteve/miniconda3/envs/conda-lock/bin/conda-lock", line 8, in <module>
    sys.exit(main())
  File "/home/local/lesteve/miniconda3/envs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
[...]

PR #328, opened 2023-02-04
Store lockfile-relative sources in lockfiles

riccardoporreca:
### Description

This is the revised fix for #229 following the original PR #230, especially given what done in #204 for the inputs metadata.

In fact, we use consistently what introduced in #204 for the actual `metadata.sources` in the lockfile, where lockfile-relative paths are always used (as the most sensible approach) unless this is made impossible by paths being on different drives, thus addressing the needs of #175, in a ways even more specific than what originally though in [#228](https://github.com/conda/conda-lock/issues/229#issuecomment-1209933411)


riccardoporreca:
I introduced a test for the lockfile-relative sources. I am not sure how we can test the case where paths are on different drives, which is now the only case when absolute paths are stored. Happy to take any suggestion, perhaps mocking `conda_lock.common.relative_path` to raise the exception thus forcing the behavior?

riccardoporreca:
There is an issue with the pre-commit, which I also had locally and seems completely unrelated

riccardoporreca:
There is also a failing `test_run_lock_regression_gh155[micromamba]` (on Python 3.8 only), which seems pretty unrelated as it should be very specific to #155

maresb:
Ya, the precommit failure is related to a bad release of isort, and is fixed by #327. For the CI failures they are random and I am rerunning the failed tests. I am trying to track down the root cause but unfortunately my time is very limited over the next few weeks.

riccardoporreca:
@maresb, I noticed the merge fro `main` has also included dropping support for python < 3.8 (#330).

There are just a few conditions like  `if sys.version_info < (3, 8)` in `test_conda_lock.py`, one of which is part of the test I introduced in the PR: I am happy to clean this up if you agree, otherwise it might be worth a second PR after this one is merged

maresb:
Ya, I think it's fine to wait, especially since the cases you mention are so clearly marked. If someone comes along with an objection before the next release I am open to reverting #330, so I am glad to retain 3.6 compatibility for now.

At some point we should modernize the coding style, like by adding pyupgrade as a pre-commit hook. But before that I am hoping that we can wind down the review queue enough in order to avoid creating nasty merge conflicts for the current contributors.

maresb (approved):
Thanks a lot for your patience and persistence on this. @mariusvniekerk, could you please take a look?

mariusvniekerk (approved):

maresb:
Thank you @riccardoporreca!!!

PR #330, opened 2023-02-05
Drop support for Python 3.6 and 3.7

### LINKED ISSUE #329: Drop support for Python 3.6 and 3.7
### Checklist

### What is the idea?

Ease development through use of modern Python

### Why is this needed?

Working with type annotations becomes much easier when I can use `from __future__ import annotations`

### What should happen?

Bump minimal Python version to 3.7

### Additional Context

Any thoughts, @bollwyvl?

maresb:
Closes #329


pyproject.toml
```diff
@@ -10,16 +10,14 @@ authors = [
...
     "Intended Audience :: Science/Research",
     "Operating System :: OS Independent",
     "License :: OSI Approved :: MIT License",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.6",
-    "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
```
srilman, line 23: Should we add 3.11 here?

maresb: Yes, good catch, thanks!!!

Issue #338, opened 2023-02-09
Flaky CI tests: KeyError: 'md5'


maresb:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

In the tests for PRs many jobs are failing with `KeyError: 'md5'` in the [`_reconstruct_fetch_actions` function](https://github.com/conda/conda-lock/blob/d04563cd8f2001e64316eb46975781bcda29d9ce/conda_lock/conda_solver.py#L277).

Rerunning the failed jobs typically fixes the error, but it still causes a lot of confusion.

~I have also seen a less common error: `RuntimeError: No user found` in `setup of test_private_lock[micromamba-${QUETZ_API_KEY}]` from our [Quetz server test fixture](https://github.com/conda/conda-lock/blob/d04563cd8f2001e64316eb46975781bcda29d9ce/tests/conftest.py#L63).~ :heavy_check_mark: 

We should look into this and fix it, and if you're submitting a PR, it's not your fault!

maresb:
The Quetz API key seems to be particularly problematic at the moment.

Locally I try:

```
cd tests
docker run --rm -it -v $PWD/quetz/dev_config.toml:/etc/config.toml:ro mambaorg/quetz quetz run --copy-conf /etc/config.toml --dev --host 0.0.0.0 /run/quetz
```

and then I see a matching line like

```
Test API key created for user "carol": 2069ab85c2b3490ca4d3dc3f339144f4
```

so it's not immediately clear what goes wrong. Maybe we should dump the logstream before failing so that we can look for particular errors.

maresb:
I did a quick experiment in #348 to try and get some more info about the md5 key error.

It seems to have to do with the data you get when running a command like:

```bash
docker run --rm -it mambaorg/micromamba:1.3.1
micromamba create --prefix /tmp/prefix --dry-run --json -v --channel=conda-forge zlib
```

This generates a dictionary where `actions.FETCH` and `actions.LINK` are lists of 6 elements, a typical element looking like

```json
{
    "build": "h166bdaf_4",
    "build_number": 4,
    "build_string": "h166bdaf_4",
    "channel": "https://conda.anaconda.org/conda-forge/linux-64",
    "constrains": null,
    "depends": [
        "libgcc-ng >=12",
        "libzlib 1.2.13 h166bdaf_4"
    ],
    "fn": "zlib-1.2.13-h166bdaf_4.tar.bz2",
[...]

PR #340, opened 2023-02-09
Remove all References of `pip_support` Flag

### LINKED ISSUE #339: Remove `PIP_SUPPORT` Flag from Codebase
### Checklist

### What is the idea?

In order to get rid of the `Poetry` dependency when specifying pip packages, we created a vendored copy of the `poetry` repo inside of `conda-lock`. However, we still have this concept of enabling and disabling `PIP_SUPPORT` in the codebase itself. For example, when importing the `conda_lock.pypi_solver` module, we do the following:
```
try:
    from conda_lock.pypi_solver import solve_pypi

    PIP_SUPPORT = True
except ImportError:
    PIP_SUPPORT = False
```
However, it is unlikely for this import to fail since the `pypi_solver` module now only imports required dependencies.

Furthemore, we still have references to the `pip_support` extra in our GitHub README and docs, despite not supporting that extra anymore:
> ### pip support
>`conda-lock` can also lock the `dependencies.pip` section of
> [environment.yml][envyaml], using [Poetry's][poetry] dependency solver, if
> installed with the `pip_support` extra.

We should remove all references of `PIP_SUPPORT` in our codebase for simplicity and clarity.

### Why is this needed?

It would make the codebase a little bit simpler and clearer for new users and contributers.

### What should happen?

There shouldn't be any change from a user's perspective, other than the documentation not referencing the `pip_support` extra anymore.

### Additional Context

_No response_

srilman:
### Description
Closes #339 by removing any references of the `pip_support` extra from the docs and README. It also removes the vacuous use of the PIP_SUPPORT flag and anywhere that flag is passed into, especially in the `conda_lock.src_parser.environment_yaml.py` module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,7 @@ The default category is `main`.
 ### pip support
 
 `conda-lock` can also lock the `dependencies.pip` section of
-[environment.yml][envyaml], using [Poetry's][poetry] dependency solver, if
-installed with the `pip_support` extra.
+[environment.yml][envyaml], using a vendored copy of [Poetry's][poetry] dependency solver.
 
 ### private pip repositories
 Right now `conda-lock` only supports [legacy](https://warehouse.pypa.io/api-reference/legacy.html) pypi repos with basic auth. Most self-hosted repositories like Nexus, Artifactory etc. use this. To use this feature, add your private repo into Poetry's config _including_ the basic auth in the url:
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -53,15 +53,6 @@ from conda_lock.invoke_conda import (
     determine_conda_executable,
     is_micromamba,
 )
-from conda_lock.models.channel import Channel
-
-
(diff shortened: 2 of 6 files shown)


srilman:
I'm seeing test cases that use Docker Quetz error out at the setup stage:
```
ERROR at setup of test_private_lock[conda-$QUETZ_API_KEY]
...
```
I don't see anything in `conftest.py` that indicates that this PR was the cause of the issue, so I will rerun CI in ~1 hour or so.

srilman:
Thanks @maresb!

srilman:
Hmm, I'm still seeing the Quetz tests erroring out in the pytest fixture `quetz_server`

maresb:
Regarding Quetz failures see https://github.com/conda/conda-lock/issues/338. I think we just keep retrying for now. :slightly_frowning_face:

maresb:
Thanks @srilman!!!

Issue #315, opened 2023-01-14
Hanging on my Local Environment when including PyPi dependencies


srilman:
I have noticed that recently, `conda-lock` seems to hang on my local machine when building a lockfile for a specification that includes pip dependencies. I do not see any hangs when building conda-only lockfiles. And when I tried conda-lock in the past a couple of months ago, I did not see this problem.

For example, I tested building a lockfile for the following environment file
```
channels:
  - conda-forge

platforms:
  - linux-64

dependencies:
  - pip
  - pip:
    - deltalake
```
I ran it using the following command
```
pipx run conda-lock -f ex.yml --log-level DEBUG    
```
I also tested the `main` branch version by running 
```
pipx run --spec "git+https://github.com/conda-incubator/conda-lock.git@main" conda-lock -f ex.yml --log-level DEBUG
```

When I tested on a docker container, building the lockfile took under a minute. When testing on the base machine (which is a Mac running on Python 3.10 installed from Homebrew), it was still running even after 10 minutes. This was the last log output
```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/deltalake/json HTTP/1.1" 304 0
DEBUG:conda_lock._vendor.poetry.repositories.pypi_repository:<debug>PyPI:</debug> 31 packages found for deltalake *
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/pyarrow/json HTTP/1.1" 304 0
DEBUG:conda_lock._vendor.poetry.repositories.pypi_repository:<debug>PyPI:</debug> No release information found for pyarrow-0.1.0, skipping
DEBUG:conda_lock._vendor.poetry.repositories.pypi_repository:<debug>PyPI:</debug> 5 packages found for pyarrow >=7
DEBUG:conda_lock._vendor.poetry.repositories.pypi_repository:<debug>PyPI:</debug> The cache for pyarrow 10.0.1 is outdated. Refreshing.
DEBUG:conda_lock._vendor.poetry.repositories.pypi_repository:<debug>PyPI:</debug> Getting info for pyarrow (10.0.1) from PyPI
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
[...]

PR #342, opened 2023-02-10
Have Vendored Poetry Store Config, Cache, and Data Folders in "pypoetry-conda-lock"

srilman:
### Description
We determined that the problems we noticed in #315 was caused when Poetry is installed on a user's machine and it's cache conflicts with the vendored poetry's cache. By modifying the locations inside of vendored poetry, we can ensure that it
- Stores it cache, config files, and data folder in a different folder from poetry. Poetry stores them in a folder called "pypoetry" and we will in "pypoetry-conda-lock". Note that the location of the folder is OS specific. More details can be found here: https://python-poetry.org/docs/configuration/#default-directories
- Ensure that vendored poetry will not use any config options a user sets for regular poetry

Note that as @maresb suggested in https://github.com/conda/conda-lock/issues/315#issuecomment-1418361666, I added a option to `vendoring` in `pyproject.toml` to replace any instances of `"pypoetry"`, but manually modified the affected file myself since re-vendoring Poetry is not possible right now.

### DIFF
--- a/conda_lock/_vendor/poetry/locations.py
+++ b/conda_lock/_vendor/poetry/locations.py
@@ -6,8 +6,8 @@ from .utils.appdirs import user_config_dir
 from .utils.appdirs import user_data_dir
 
 
-CACHE_DIR = user_cache_dir("pypoetry")
-CONFIG_DIR = user_config_dir("pypoetry")
+CACHE_DIR = user_cache_dir("pypoetry-conda-lock")
+CONFIG_DIR = user_config_dir("pypoetry-conda-lock")
 
 REPOSITORY_CACHE_DIR = Path(CACHE_DIR) / "cache" / "repositories"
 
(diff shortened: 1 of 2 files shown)


maresb (approved):

maresb:
In preparation of merging I removed the debug commit and rebased on main.

maresb:
I will be entirely AFK a while, but I would have liked to merge as soon as this turns green. Perhaps we should enable GH automerge in this repo.

srilman:
Why not allow the PR creator to merge their PR when ready? We can set GitHub up to only allow Merges (not Squash or Rebase) and to only enable it once all checks are passing.

maresb:
Thanks a lot @srilman for all your help!!!

> Why not allow the PR creator to merge their PR when ready? We can set GitHub up to only allow Merges (not Squash or Rebase) and to only enable it once all checks are passing.

Do you know the specific settings we'd need to change to set this up?

PR #352, opened 2023-02-15
Invoke codecov only in CI

### LINKED ISSUE #351: Codecov seems to break interactive debugging
### Checklist

### What happened?

Since the merge of #333 I've been having trouble using the VS Code interactive debugger. I think we're running into https://github.com/microsoft/vscode-python/issues/693.

I like having the settings consolidated in pyproject.toml, but would it make more sense in this case to only run codecov in CI? 

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

maresb:
Closes #351

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
           pushd "${RUNNER_TEMP}"
           set TMPDIR="%RUNNER_TEMP%"
           dir
-          pytest tests
+          pytest --cov=conda_lock --cov-branch --cov-report=xml --cov-report=term tests
           copy coverage.xml %GITHUB_WORKSPACE%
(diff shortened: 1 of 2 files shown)


maresb:
Hi @dbast, could you please take a look at issue #351 when you get the chance? For some reason, codecov seems to break interactive debugging in VS Code and more. Have you run into this before? Does this PR look like a sensible fix, or do you have another suggestion?

dbast (approved):

maresb:
Ya, I thought about that comment, but it seems to me like it's VS Code specific solution, while the issue affects other IDEs as well. 

Thanks a lot for the quick approval! I'll merge soon, but I'm happy to revisit this if another solution emerges.

Issue #350, opened 2023-02-14
Step-By-Step Documentation


jesshart:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

I am working on a step-by-step how-to document for `conda-lock` and I would like to donate all the time I've spent on it to the official documentation.

### Why is this needed?

The existing documentation "gets the job done," but I would like to provide something for devs like me who are slower to get the idea how how a tool works.

### What should happen?

I have already done a good deal of work testing my tutorial. My request would be that the community could help me review and we could publish it to the repo.

maresb:
Hey, this sounds excellent!!! Do you have a draft PR we could look at?

jesshart:
> Hey, this sounds excellent!!! Do you have a draft PR we could look at?

I love the enthusiasm! I will get a PR together asap. I have contributed a ton on teams using Bitbucket (i.e., elsewhere) so bare with me as I learn the ropes here.

jesshart:
@maresb reviewed `CONTRIBUTING.md` and got up to the `pytest` command but I am unsure if I should proceed.

```bash
25 failed, 84 passed, 6 errors in 129.63s (0:02:09)
```

Is this expected or do I need to try again? I am using `conda` with `libmamba` which is not what the contributor docs call for.

maresb:
If you're just doing doc-related stuff, don't worry so much about pytest. (It will anyways automatically run remotely in the CI after you push.)

If you want to debug pytest, then I'll need some details on the error messages.

jesshart:
Alright! I forked, cloned, did the work, pushed it, and then requested a PR. I hope that was the right order of things.

Issue #158, opened 2022-03-01
[Feature Request] use existing lock-file to optimize solver


nkrishnan:
A common use-case for conda-lock is when modifying an existing package dependency spec and regenerating the lock files. For this use case is it possible to add support for an "incremental" mode where we use the existing lock file as the answer of first resort to short circuit a more expensive search? And for users to optionally invoke via `conda-lock lock --bootstrap-lock-file <lockfile>` or similar.

mariusvniekerk:
I think with some of the new features that are coming with the updatable lockfiles that should be doable.  

The best solution for performance atm is to always ensure that you are running with 

```
conda-lock --mamba
```

This will become the default behavior shortly (if mamba is detected on PATH)

wolfv:
For mamba, I've been experimenting with generating a transaction from an initial solve, and then use that transaction to create "fake installed" packages, and then continue modifying the transaction here: https://github.com/mamba-org/mamba/pulls

It's not done yet, but the goal is to support also a use case where you make an environment larger and then subtract packages you don't actually need (to make sure that the whole env is always consistent).

Maybe this could be useful in this use case as well, though.

PR #355, opened 2023-02-18
create file with draft pasted

jesshart:
### Description
This is the draft walkthrough I mentioned on issue [#350](https://github.com/conda/conda-lock/issues/350).

### DIFF
--- a/docs/walkthrough.md
+++ b/docs/walkthrough.md
@@ -1,0 +1,320 @@
+
+# Table of Contents
+- [How to Use This Guide](#how-to-use-this-guide)
+  - [Gotchas](#gotchas)
+- [Goals for the Environment](#goals-for-the-environment)
+  - [Q&A](#qa)
+- [Your Platform Setup](#your-platform-setup)
+  - [Global Dependencies](#global-dependencies)
+- [Workflow 1: Locking a New Project](#workflow-1-locking-a-new-project)
+- [Workflow 2: Locking an Existing Project (TBD)](#workflow-2-locking-an-existing-project)
+- [Reproduce](#reproduce)
+  - [With Docker in Mind](#with-docker-in-mind)
+- [Upgrade](#upgrade)
+  - [Upgrade an existing package](#upgrade-an-existing-package)
+  - [Adding a New Package via `conda`](#adding-a-new-package-via-conda)
+  - [Adding a New Package via `pip`](#adding-a-new-package-via-pip)
+  - [A Quick Note on Downgrading](#a-quick-note-on-downgrading)
+- [Debugging](#debugging)
+  - [`conda-lock` giving vague `AssertionError`](#conda-lock-giving-vague-assertionerror)
+- [Terms](#terms)
+- [References and Resources](#references-and-resources)
+- [Versions](#versions)
+
+# How to Use This Guide
+This is a guide on how to produce a reproducible and upgradeable environment for Python applications. A great deal of inspiration came from a pythonspeed.com article about this very topic.<sup>[4](#references-and-resources)</sup> While the article is a wonderful resource, I needed more so I developed this guide.
+
+You can use the table of contents to get a quick look at topics. Or you can skip ahead to [Your Platform Setup](#your-platform-setup) which is the beginning of the walkthrough.
+
+Ultimately, this guide is just that, a *guide*. The goal was to make the concepts easy to follow so that you can apply it to your use case. You may use a different platform than this guide but hopefully still benefit from the content.
+
+## Gotchas
+* I use `pipx` for parts of this guide, you may choose not to.
+* My platform may not be your platform so keep an eye out for incompatible commands.
+* For simplicity, the guide assumes no explicit versions of packages.
+# Goals for the Environment
+**Reproducible**: The environment can be reproduced across platforms.
+**Upgradeable**: Packages in the environment can be easily changed (e.g., version, new package, etc.)
+
+## Q&A
+
+### Why do you suggest using  `conda`  AND  `pip`?
+Sometimes a package installed via `conda` (even using the `conda-forge` channel) breaks my environment. I have found using `pip` as the *exception to the rule* has resolved most of these issues.
+
+### Why is this process slow?
+Solving the dependencies of dependencies for one operating system is time consuming. Out of the box,  `conda-lock`  will try to solve for 3 different operating systems (i.e., linux-64, osx-64, and win-64). However, `conda` recently announced partnering with a `mamba` solver. You could configure your `conda` instance to use a faster solver like this one. This is beyond the scope of this walkthrough. Regardless, it will still take time.
+
+### Where should I install  `conda-lock`?
+Using `pipx` to install `conda-lock` allows you to have peace of mind about running the package outside of your `conda` environment (even your `conda` "base" environment). Alternatively, you could install it in your `conda` "base," but I ran into issues that way and this walkthrough will not cover that angle.
+
+### How can I find the version of my conda dependency?
+A command I like to use is,
+```bash
+(base) $ conda list | grep PACKAGE_NAME
+```
+This should show you the package with the version.
+
+# Your Platform Setup
+"Your *platform*" is the platform being used by you from a global perspective. Your environment is going to be the place we initialize the process of standardizing the environment for the goals listed above. By the end of this guide, you should be able to share the environment to another platform.
+
+## Global Dependencies
+- [pipx](https://pypa.github.io/pipx/) — "...help you install and run end-user applications written in Python"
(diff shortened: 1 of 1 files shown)


maresb:
Thanks a lot for this!!! Quickly skimming I really like the concept. I could see this turning into some sort of quickstart guide. I think people will have a substantial amount of feedback and suggestions, but working through those may be very valuable for bridging the gap between new and experienced users.

I am also not sure if/how exactly we will publish this. That is dependent on how this evolves, and also a broader consensus among the maintainers, so I can't make any promises. Also, our review capacity is a bit strained at the moment, so I really hope that spectators step up and offer their feedback.

Thanks a lot for all the careful work and thought you have put into this!

maresb:
You should probably get pre-commit running locally on your side, it is simpler than pytest.

maresb:
pre-commit.ci autofix

maresb:
Make sure to pull that pre-commit formatting fix before making additional commits in order to prevent merge conflicts.

docs/walkthrough.md
```diff
@@ -0,0 +1,320 @@
...
+3. Update the file using `conda-lock`
+```bash
+(base) $ conda-lock --update matplotlib
+Locking dependencies for ['linux-64', 'osx-64', 'osx-arm64', 'win-64']...
+...
+```
+Note: If you update to a specific version of a package, you will need to provide the version in the command. For example, `matplotlib=3.5` should be included after the `--update` flag. This update flag is useful for the following:
+* Updating to the latest version of a package (no version specified)
+* Updating a package to a specific version
+* Adding a new package
```
timsnyder-siv, line 241: Does this actually work?  I am trying to do this same type of thing.  In my case, I want to add `fsspec` to https://github.com/firesim/firesim/blob/main/conda-reqs.conda-lock.yml.  To do that, I:
1. add `fsspec` to https://github.com/firesim/firesim/blob/main/conda-reqs.yaml
2. `conda-lock -f conda-reqs.yaml -p linux-64 --lockfile conda-reqs.conda-lock.yml --update fsspec`

NOTE: I don't necessarily install the package into the conda environment because conda-lock doesn't use the contents of the currently active environment  in  it's calculation of the lockfile, does it?

However, I find that unless package you're trying to add is in the lockfile, the `intersection()` at https://github.com/conda/conda-lock/blob/854fca9923faae95dc2ddd1633d26fd6b8c2a82d/conda_lock/conda_solver.py#L151 filters the added package out of `to_update` and `conda-lock` effectively re-solves the entire environment, it doesn't use the `update` code.

Also, I think it is mildly confusing to suggest step 1 is installing the package into the currently active environment because it implies that having the packages installed somehow influences how conda-lock creates the lockfile, which I don't think is correct.  If you are using a lockfile to reproducibly create an environment, it would probably be better to create the updated lockfile and then use it to create the environment.  I haven't tried installing on top of an existing environment to see what happens but I'd expect it to be the same as removing the environment completely and reinstalling everything in the lockfile (but perhaps with optimizations for leaving unchanged packages alone).

timsnyder-siv: The `INFO` logger line after `Locking dependencies for ...` is indicative of whether update is happening or not.

When update is being used, it should say something like:
```
INFO:conda_lock.conda_solver:to_update: ['matplotlib']
```

I suspect the output looks like :
```
INFO:conda_lock.conda_solver:linux-64 using specs ['python', 'pandas', 'matplotlib']
```
and that indicates that the entire environment is being resolved and `--update` is effectively being ignored.

timsnyder-siv: @mariusvniekerk, reading https://github.com/conda/conda-lock/issues/158 it seems like incremental updating should be possible.  If you could take a look at this use-case and comment on what I've found, it would be really helpful.

jesshart: > 

I can confirm that what is shown in the walkthrough is in fact `..using specs...` and appears to be ignoring the `--update` flag. I can be sure to remove this section, or at least clarify that the environment is being solved from scratch.

maresb: I'm not very familiar with the `--update` flag, but if there's a suspected bug (or a missing warning message), let's open an issue for follow-up.

jesshart: @timsnyder-siv do you want to investigate this further and open an issue for it? It looks like #158 does not mention the `--update` flag. I see mentions of `mamba` and "performance" but I think what we are discussing might be different?

mariusvniekerk: In general the update flag should only really be used if you absolutely have to minimize the change to a lock.  The more standard approach is to just delete and get a new solve

timsnyder-siv: Thanks @mariusvniekerk.  I guess I was asking whether it is supposed to be possible to add new specs to the solve and ask to only update those new specs or if one currently must re-solve the whole environment.

mariusvniekerk: That should work i think, but yeah, probably worth testing if it behaves in the way that you'd expect

docs/walkthrough.md
```diff
@@ -0,0 +1,320 @@
...
+**Reproducible**: The environment can be reproduced across platforms.
+**Upgradeable**: Packages in the environment can be easily changed (e.g., version, new package, etc.)
+
+## Q&A
+
+### Why do you suggest using  `conda`  AND  `pip`?
+Sometimes a package installed via `conda` (even using the `conda-forge` channel) breaks my environment. I have found using `pip` as the *exception to the rule* has resolved most of these issues.
+
+### Why is this process slow?
+Solving the dependencies of dependencies for one operating system is time consuming. Out of the box,  `conda-lock`  will try to solve for 3 different operating systems (i.e., linux-64, osx-64, and win-64). However, `conda` recently announced partnering with a `mamba` solver. You could configure your `conda` instance to use a faster solver like this one. This is beyond the scope of this walkthrough. Regardless, it will still take time.
```
mariusvniekerk, line 45: `conda-lock` defaults to using mamba if its installed so we should just recommend that.

jesshart: I have been using `miniconda` with the `libmamba` solver enabled ([ref](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community)). Is that much different from using `mamba` directly?

docs/walkthrough.md
jesshart: > The more standard approach is to just delete and get a new solve

I can modify the document to recommend this approach. However, to your point,

> That should work i think, but yeah, probably worth testing if it behaves in the way that you'd expect

It appears the `--update` flag is not behaving as expected. For example, when I used it to add `matplotlib` it behaved like this:

```bash
(base) $ conda-lock --update matplotlib
INFO:conda_lock.conda_solver:linux-64 using specs ['python', 'pandas', 'matplotlib']
``` 

This is in contrast to what was expected according to @timsnyder-siv:

```bash
INFO:conda_lock.conda_solver:to_update: ['matplotlib']
```

So it appears the command is ignoring the flag.

timsnyder-siv: @jesshart we should probably open a new issue referencing this.  If we're feeling strong in our python-fu, PR'ing a test would also be helpful and if we're feeling particularly masterful, possibly a patch that helps conda-lock do this solve.  I looked at the code enough to think I might be able to come up with a patch but it will probably take me longer to get right than I have time for right now.  I'll open a ticket from this thread.

docs/walkthrough.md
mariusvniekerk: they're similar but not the same

jesshart: Not trying to be obtuse here, but wouldn't a strong dependency like this discourage a variety of users from engaging `conda-lock`? I am happy to adapt this documentation, but wanted to approach this philosophical point first.

For example, I could make the point in the documentation so readers are not misled into thinking it only works with the `mamba` setup.

jesshart:
Document Changes
- [ ] Replace `libmamba` approach with  a `mamba` approach ([different from utilizing `libmamba` in conda](https://github.com/conda/conda-lock/pull/355#discussion_r1117580303))
- [ ] Remove references to `--update` flag since it is not working as expected (#386)

jesshart:
@maresb I updated the document per the conversations here. However, I forgot to follow your instructions,

> Make sure to pull that pre-commit formatting fix before making additional commits in order to prevent merge conflicts.

Which may have caused a problem. I resolved the merge conflicts on my end, but it looks like they still broke during the `pre-commit.ci - pr` step. Do I need to do anything on my end?

maresb:
@jesshart, if you like your local version, then the easiest solution is to do a force-push `git push --force` and this will replace the conflicting `walkthrough-doc` branch here with your local copy.

mariusvniekerk:
pre-commit.ci autofix

Issue #334, opened 2023-02-07
Default PyPI dependencies for pyproject.toml


pep-sanwer:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

My understanding is that as `pyproject.toml` (Poetry flavor) support currently stands, all dependencies default to conda channels unless explicitly stated as `source = "pypi"` for that particular dependency. From what I understand, this includes dependencies listed in `[tool.poetry.dependencies]` as well as `[tool.conda-lock.dependencies]`. 

The feature suggestion here is change this default behavior, so that dependencies listed in `[tool.poetry.dependencies]` default to PyPI, i.e. `source = "pypi"` without having to explicitly state it in `pyproject.toml`, while `[tool.conda-lock.dependencies]` behavior can remain the same, i.e. defaulting to conda channels unless explicitly stated otherwise in `pyproject.toml`.

This would enable workflows where you can `conda-lock` to solve an environment, `conda-lock install` to install a reproducible environment, and then `poetry add pandas` (for example) to add additional pyPI dependencies as the project evolves. As it currently stands, one would have to invoke `poetry add pandas --source="pypi"` (for example) per additional pyPI dependency.

Currently, the workflow I have landed upon is to lock and reproduce the environment with:
- `conda-lock --conda mamba -f pyproject.toml`
- `conda-lock install --conda mamba --name test_env conda-lock.yml`
- then an additional `poetry install` to update all pyPI packages from conda channels to their latest pyPI
- my concern is that that the previous step may cause package issues (I've been burned in the past mixing conda and pip installs in the same project env)

If I am misunderstanding how additional dependencies should be added to a project using `conda-lock` & `pyproject.toml` or how conda-lock should be utilized with Poetry, please correct me.

An additional question would also be, does `conda-lock` default to `conda-lock --conda mamba` if the only `conda` installed is from `mamba-forge`?

Appreciative of this wonderful tool!

maresb:
Thanks a lot @pep-sanwer for the thoughtful comments! I like your suggestion, and your explanation of your usage. To put it in my own words:

* PyPI dependencies should go in `[tool.poetry.dependencies]`.
* Conda dependencies should go in `[tool.conda-lock.dependencies]`.

[...]

PR #362, opened 2023-02-23
Default PyPI dependencies for pyproject.toml #334

pep-sanwer:
Quick draft implementation of changes discussed in #334:

Implements a `default-poetry-source-pypi` flag in `[tool.conda-lock]` to default to `source = "pypi"` for all dependencies in `[tool.poetry.dependencies]`, i.e.:
- Defaulting to PyPI dependencies for `[tool.poetry.dependencies]`
- Defaulting to Conda dependencies for `[tool.conda-lock.dependencies]`

by explicitly providing  `default-poetry-source-pypi = true` in `[tool.conda-lock]` section, e.g.:
```toml
[tool.conda-lock]
default-poetry-source-pypi = true
```

Updated Readme and tests to reflect changes. Please let me know any thoughts or comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,8 +388,18 @@ A dependency will also be treated as a `pip` dependency if explicitly marked wit
 [tool.conda-lock.dependencies]
 ampel-ztf = {source = "pypi"}
 ```
+##### Defaulting Poetry Dependency Source to PyPI
 
-In both these cases, the dependencies of `pip`-installable packages will also be
+Alternatively, the above behavior is defaulted for all dependencies defined in `[tool.poetry.dependencies]`, i.e.:
+- Defaulting to PyPI dependencies for `[tool.poetry.dependencies]`
+- Defaulting to Conda dependencies for `[tool.conda-lock.dependencies]`
+
+by explicitly providing  `default-poetry-source-pypi = true` in `[tool.conda-lock]` section, e.g.:
+```toml
+[tool.conda-lock]
+default-poetry-source-pypi = true
+```
+In all cases, the dependencies of `pip`-installable packages will also be
 installed with `pip`, unless they were already requested by a `conda`
 dependency.
 
--- a/README.md
+++ b/README.md
@@ -394,10 +394,10 @@ Alternatively, the above behavior is defaulted for all dependencies defined in `
 - Defaulting to PyPI dependencies for `[tool.poetry.dependencies]`
 - Defaulting to Conda dependencies for `[tool.conda-lock.dependencies]`
 
-by explicitly providing  `default-poetry-source-pypi = true` in `[tool.conda-lock]` section, e.g.:
+by explicitly providing  `default-dependencies-to-conda = false` in `[tool.conda-lock.poetry]` section, e.g.:
 ```toml
-[tool.conda-lock]
-default-poetry-source-pypi = true
+[tool.conda-lock.poetry]
+default-dependencies-to-conda = false
 ```
 In all cases, the dependencies of `pip`-installable packages will also be
 installed with `pip`, unless they were already requested by a `conda`
--- a/README.md
+++ b/README.md
@@ -388,17 +388,18 @@ A dependency will also be treated as a `pip` dependency if explicitly marked wit
 [tool.conda-lock.dependencies]
 ampel-ztf = {source = "pypi"}
 ```
-##### Defaulting Poetry Dependency Source to PyPI
+##### Defaulting non-conda dependency sources to PyPI
 
-Alternatively, the above behavior is defaulted for all dependencies defined in `[tool.poetry.dependencies]`, i.e.:
-- Defaulting to PyPI dependencies for `[tool.poetry.dependencies]`
-- Defaulting to Conda dependencies for `[tool.conda-lock.dependencies]`
+Alternatively, the above behavior is defaulted for all dependencies defined outside of `[tool.conda-lock.dependencies]`, i.e.:
+- Default to `pip` dependencies for `[tool.poetry.dependencies]`, `[project.dependencies]`, etc.
+- Default to `conda` dependencies for `[tool.conda-lock.dependencies]`
 
-by explicitly providing  `default-dependencies-to-conda = false` in `[tool.conda-lock.poetry]` section, e.g.:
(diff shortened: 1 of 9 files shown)


maresb:
Thanks @pep-sanwer, this looks good.

Could you please remove the references to Poetry? Simply because specifying dependencies in `pyproject.toml` is not Poetry-specific. (All standards-compliant build tools specify their dependencies in `pyproject.toml`, not just Poetry.) The [standard location](https://peps.python.org/pep-0621/#dependencies-optional-dependencies) for specifying dependencies in `pyproject.toml` is `[project.dependencies]`, used by all major build backends except for Poetry (when dependencies aren't [dynamically]() specified).

Unfortunately my alternative suggestions for wording and variable names at the moment are a bit awkward... I'd say that dependencies specified in `pyproject.toml` are "project dependencies" or `pyproject.toml` dependencies. Maybe call the variable `default-pyproject-source`?

Finally, I think we may be limiting ourselves by having the setting be a boolean. Namely, we may want to specify a channel, or a list of channels. How about something like:

```
dependency-source-priority = ["pypi", "conda-forge"]
```

@croth1, could you please have a look at this? (He has been thinking about channels lately.)

croth1:
With limited knowledge of the whole system here are my two cents:

I like the idea of ultimately having something like: `dependency-source-priority = ["pypi", "conda-forge"]` - the way we now approach python dependencies as a two-stage solving process, would probably make it incredibly hard to allow mixing pypi and conda channels in the dependency declaration (e.g. ["private-pypi", "pytorch", "conda-forge", "pypi"]. So naturally I feel that we might converge towards separate: `pypi-source-priority = ["pypi", "conda-forge"]` and conda-channels. However this would still require the boolean choice which one comes first.

(side note: I find it extremely attractive to explore whether instead of solving with poetry ontop of conda, and then trying to mingle everything together, we could make a fake-repo just like for the virtual packages and solve everything together with the conda solver. While this seems technically much harder, I have a strong suspicion that this will be conceptually much easier and a lot of headaches we have right now might just magically disappear)

Back to the topic 😬 - as I see it the changes in this PR are completely poetry-toml specific and I feel it's worth offering this, without having it to fit into a grand vision. For now, I would suggest to keep the changes as they are and configure in a poetry specific namespace:

```toml
[tool.conda-lock.pypoetry_toml]
default-dependencies-to-conda = false
```

Should be easy to deprecate and pull these configurations into the main `[tool.conda-lock]` section once we have source-agnostic general solution. What do think?

maresb:
Thanks a lot @croth1 for the feedback!

I agree that the "right" way to do things would be with a unified solver. I'm very interested in looking into what this would involve, and I suspect we could use something like [libsolv](https://mamba.readthedocs.io/en/latest/developer_zone/internals.html). It might be interesting to read up on the details and ping wolfv for advice. But unless some expert like wolfv steps in, I don't see this happening anytime soon.

As for a short/medium-term solution, based on @croth1's advice, I think it makes sense that we specify which package manager (pip vs conda) to prefer by default.

As for terminology, we already have this in the new-style unified lockfile. There we use a categorical value `package[].manager` with values `conda` or `pip`. While this is a binary choice, I don't see it as a "boolean" true/false choice, so I think strings would be better here. For example:

```toml
[tool.conda-lock]
default-project-dependency-manager = "pip"
```

Now that we have identified "dependency managers" rather than "channels" as the appropriate level of abstraction, I wonder again if it might make sense to use a list here, so something like

```toml
dependency-manager-priority = ["conda", "pip"]  # the current default
dependency-manager-priority = ["pip"]  # pip-only
dependency-manager-priority = ["pip", "conda"]  # pip with conda as fallback
dependency-manager-priority = ["conda"]  # strict conda mode
```

Do you think this would be useful? Or more importantly doable?

### Terminology rant...

In short, in order to avoid perpetuation of confusion, I simply want to use the proper terminology. Specifically, "dependencies specified in `pyproject.toml`" are not "Poetry dependencies" in the same sense that "pain-relief drugs" are not "Tylenol".

I've seen widespread confusion (and been confused myself) about the Python packaging ecosystem, and a big part of this appears to be [generization](https://en.wikipedia.org/wiki/Generic_trademark) which falsely conflates various tools and standards. (For example, I used to conflate pip with setuptools.)

I get concerned when I see "poetry-toml", because this serves to perpetuate confusion that `pyproject.toml` is a Poetry file. All modern Python projects should strive to declare their dependencies in `pyproject.toml`, whether or not they use Poetry. Poetry is not required in order to use a `pyproject.toml`. Moreover, conda-lock already supports non-Poetry `pyproject.toml` usage.

For a concrete example, conda-lock doesn't use Poetry, and conda-lock can generate lockfiles [for itself](https://github.com/conda/conda-lock/pull/359) (up to a few remaining general issues).

maresb:
What's the expected behavior for non-Poetry `pyproject.toml` files, i.e. when dependencies are defined in `[project.dependencies]`?

maresb:
This looks great!!! Thanks so much for the update.

I won't be able to look at it this week unfortunately, due to time constraints.

maresb:
Thanks so much for your perseverance @pep-sanwer!

Let's do one more rebase. I'll squeeze this in before the Poetry revendor.

maresb:
The branch is out-of-date again, but don't bother rebasing this time, as it should be totally independent.

Now we just have to deal with this recent flakiness with the failing tests... :disappointed:

maresb (approved):
Thanks so much @pep-sanwer for the extra effort here to make this work for all dependency managers! Excellent work!

maresb:
Hey, thanks a lot for the report @pep-sanwer! No worries, this stuff happens. It looks like we just missed a test case. On the bright side, this is a new feature, so people aren't relying on it yet.

One quirk of Poetry is that it doesn't define the Python dependency in `project.requires-python`, so this might be a Poetry-only issue.

But with a brief glance through the code, I don't see anywhere that we're parsing `requires-python`. :see_no_evil: Maybe we're ignoring it? That's yet another issue to open.

Would you like to open a PR with your fix for Poetry plus updated tests?

maresb:
Thanks a lot @pep-sanwer for looking into this!

A comprehensive solution would be excellent, but in case it helps to make incremental progress, I think we could break this one up into the Poetry and non-Poetry parts as separate pieces. (I conceptualize this as two separate bugs.)

PR #364, opened 2023-02-25
Refactor lookup._LookupLoader

maresb:
Locally I see test failures in `test_choose_wheel` due to the assertion in `conda_lookup()` failing. It seems to result from flaky caching logic. Using `python>=3.8`'s functools, we can get rid of most of the boilerplate. In a subsequent PR we could also remove the module-level boilerplate.

In short, `_LookupLoader` was implementing a bunch of caching functionality by hand. This simply uses modern Python and `functools` to eliminate all that boilerplate.

### DIFF
--- a/conda_lock/lookup.py
+++ b/conda_lock/lookup.py
@@ -1,4 +1,5 @@
-from typing import Dict, Optional
+from functools import cached_property
+from typing import Dict
 
 import requests
 import yaml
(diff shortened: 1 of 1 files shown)


maresb:
Here is an instance in the CI of the failure I was seeing:
https://github.com/conda/conda-lock/actions/runs/4271730580/jobs/7436402534

maresb:
@mariusvniekerk this occurs an order-of-magnitude less than the md5 keyerror, but this still occasionally snags up the CI.

mariusvniekerk (approved):

PR #369, opened 2023-02-25
Use pytest-split to distribute CI tests over multiple runners

maresb:
### Description

To be squash-merged


maresb:
Here are the results of one complete run:
image

.test_durations
```diff
@@ -0,0 +1,120 @@
+{
```
mariusvniekerk, line 1: Guess this needs a gitignore

mariusvniekerk (commented):
Nice

.github/workflows/test.yml
```diff
@@ -61,6 +66,18 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest ]
         python-version: [ "3.8", "3.11" ]
+        pytest-split-group: [1, 2, 3]
```
mariusvniekerk, line 69: If we can use non integer identifiers for these groups that would be better, since I have no idea what group 4 means

maresb: I can add comments. Pytest-split tries to divide the tests into n groups labeled 1, ..., n which have approximately the same total runtime. So they have to be integers.

.test_durations
maresb: Nope, it's required, but it could probably live under tests/

.github/workflows/test.yml
maresb: The runs are labeled like "test / ... (3, 5)" which is run number 3 out of 5.

maresb:
Superseded by #689

PR #374, opened 2023-02-26
Parse Platforms Before Determining Dependencies

### LINKED ISSUE #337: Default platforms are unexpectedly added from multiple sources
### Checklist

### What happened?

If I use both `pyproject.toml` and a supplemental `environment.yml` as sources, and if I specify platforms in `pyproject.toml` but not in `environment.yml`, I would expect the resulting platforms to be the list from `pyproject.toml`. Instead it's those platforms union the default platforms (linux-64, osx-64, win-64).

The logic in [`parse_source_files`](https://github.com/conda/conda-lock/blob/d04563cd8f2001e64316eb46975781bcda29d9ce/conda_lock/conda_lock.py#L869) looks odd to me. It seems like we add the default platforms if they are missing from _any_ source file (except for `pyproject.toml`?).

I would have expected that we take the union of all platforms from all source files, and then substitute the default platforms only if that were empty.

Ping @srilman since this involves code we've been looking at recently.

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

srilman:
### Description
Closes #337. This PR changes `_parse_source_files` to first determine the list of platforms to render for by reading the source files. It will only do so if an explicit list of platforms have not been passed in.

### DIFF
--- a/conda_lock/src_parser/__init__.py
+++ b/conda_lock/src_parser/__init__.py
@@ -7,7 +7,10 @@ from conda_lock.common import ordered_union
 from conda_lock.models.channel import Channel
 from conda_lock.models.lock_spec import Dependency, LockSpecification
 from conda_lock.src_parser.aggregation import aggregate_lock_specs
-from conda_lock.src_parser.environment_yaml import parse_environment_file
+from conda_lock.src_parser.environment_yaml import (
+    parse_environment_file,
+    parse_platforms_from_env_file,
+)
 from conda_lock.src_parser.meta_yaml import parse_meta_yaml_file
 from conda_lock.src_parser.pyproject_toml import parse_pyproject_toml
 from conda_lock.virtual_package import FakeRepoData
@@ -35,9 +38,7 @@ def _parse_platforms_from_srcs(src_files: List[pathlib.Path]) -> List[str]:
         elif src_file.name == "pyproject.toml":
             all_file_platforms.append(parse_pyproject_toml(src_file).platforms)
         else:
-            all_file_platforms.append(
-                parse_environment_file(src_file, DEFAULT_PLATFORMS).platforms
-            )
+            all_file_platforms.append(parse_platforms_from_env_file(src_file))
 
     return ordered_union(all_file_platforms)
 
--- a/conda_lock/src_parser/__init__.py
+++ b/conda_lock/src_parser/__init__.py
@@ -3,6 +3,7 @@ import pathlib
 
 from typing import AbstractSet, List, Optional, Sequence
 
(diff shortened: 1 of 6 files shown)


srilman:
Tried about 3 times and keep getting the md5 error, so I'm going to retry much later.

maresb:
This is awesome!!! And thanks to your commit structure I believe I understand it.

I edited your description to change "Solves" → "Closes" since GitHub recognizes the "closes" keyword and automatically links the issue.

Sorry about the md5, I'm actually very close to solving it in #348. It seems to be some race condition where it can be overcome by repeating the dry-run solve. 

For a current workaround, I'm able to rerun individual tests after clicking "Details".

image

But maybe you don't have the same privileges. (I'm not sure what's involved in granting those...)

maresb (approved):

maresb:
The tests are green, but I force-pushed to reset your retry commits, and it restarted the tests. Now I'm blocked from merging until the required tests pass, but I'll merge tomorrow as soon as I can.

maresb:
If you get the chance it'd be really nice to clean up some of this code with a refactoring commit while it's still fresh in our minds.

The code seems fairly awkward which we might be able to clean up with something like a `SourceFile` class with a uniform interface. Also, it'd be nice if the source file type were detected by content rather than filename.

srilman:
@maresb Thanks for handling all this! 

> For a current workaround, I'm able to rerun individual tests after clicking "Details".

Yeah I think you need special privileges to do that. 

> The code seems fairly awkward which we might be able to clean up with something like a SourceFile class with a uniform interface.

Yeah I felt so too. Using a SourceFile class would make it a bit neater by parsing each file once and rendering output when necessary. This is my plan for next PRs
- Refactor LockSpecification to represent dependencies as a dictionary between the platform and list of deps. 
- Refactor to use ruamel-yaml
- Refactor to use a SourceFile class

> Also, it'd be nice if the source file type were detected by content rather than filename.

Good point. I think we can assume that the file extensions are correct (YAML files end with `.yml` or `.yaml` and TOML files end with `.toml`) then:
- pyproject-toml files are any TOML files
- meta-yaml files must be YAML files with headers `package/name` and `package/version` https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html
- environment-yaml files are all other YAML files

maresb:
Thanks a lot for the contribution. Your plan looks excellent! :rocket:

maresb:
Possibly related, though possibly just a problem with micromamba: https://github.com/mamba-org/mamba/issues/1209#issuecomment-1447546266

CC

PR #376, opened 2023-02-28
Simplify reconstruction of fetch actions

maresb:
I'm very interested to see if this has any affect on #338.

### DIFF
--- a/conda_lock/conda_solver.py
+++ b/conda_lock/conda_solver.py
@@ -22,6 +22,8 @@ from typing import (
 )
 from urllib.parse import urlsplit, urlunsplit
 
+import yaml
+
 from typing_extensions import TypedDict
 
 from conda_lock._vendor.conda.models.match_spec import MatchSpec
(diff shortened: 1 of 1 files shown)


maresb:
pre-commit.ci autofix

mariusvniekerk:
Seems like this does improve the md5 test issue?

maresb:
:point_up: squash

@mariusvniekerk, yes, this resolves it. In the process I noticed a whole bunch of illogical stuff regarding the formation of LockedDependencies which I wanted to clean up, but it goes deeper than I originally thought. I think it makes sense to merge this so that CI works, and follow up with the remaining stuff in future PRs.

:point_down: rebase

maresb:
@mariusvniekerk, all tests are green, so if you approve, then we can merge this, fixing most of the CI issues.

As a follow-up to this, we should make sure that the approaches in `update_specs_for_arch` and `solve_specs_for_arch` are as similar as possible, and eliminate duplicated code. Moreover, since the data structures for [LinkAction](https://github.com/conda/conda-lock/blob/c3b63ae2f09666e646e10185197709ea76320fbf/conda_lock/conda_solver.py#L62) are different between Micromamba and Conda/Mamba, we should redefine `LinkAction` to `Union[MicromambaLinkAction, CondaLinkAction]` with the respective models. This lights up the type checker, revealing some substantial issues, which is why I think it deserves its own PR.

mariusvniekerk (approved):

mariusvniekerk:
cc @wolfv there is probably some stuff in this discussion that you should be aware of

wolfv:
@maresb can you share a bit more of the symptoms? What I understand is that the repodata_record.json contains the correct complete values for MD5 and SHA256, but micromamba is somehow not reading them?

PR #388, opened 2023-03-13
Handle Poetry Optional Dependency Quirks 

### LINKED ISSUE #385: Optional Dependencies Have Different Behavior in Conda Lock vs Poetry
### Checklist

### What happened?

I noticed this odd combination that Poetry seems to handle in one way and conda-lock does in another. Assume we have the following `pyproject.toml`

## Situtation 1
```
[tool.poetry]
name = "conda-lock-test-poetry"
version = "0.0.1"

[tool.poetry.dependencies]
toml = ""
tomlkit = { version = "", optional = true }

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"

[tool.conda-lock]
platforms = ["osx-arm64"]
channels = ["conda-forge"]
```

This is a valid specification in poetry. However, if we were to lock and install this file with Poetry, it would only install the `toml` package and it's dependencies in the virtualenv. It seems like optional dependencies that aren't part of any extra category can never be installed by Poetry (see https://stackoverflow.com/questions/60971502/python-poetry-how-to-install-optional-dependencies).

Conda lock has slightly different behavior; it will lock both `toml` and `tomlkit` in the lockfile under the main category, but the lockfile will specify that `toml` is required and `tomlkit` is optional. However, when installing from the lockfile, `tomlkit` will be included in the resulting env by default, and I don't think there is any way to exclude it.

## Situtation 2
Furthermore, it seems like dev dependencies (specified in `tool.poetry.dev-dependencies` section) can and should never be optional (see https://stackoverflow.com/questions/60971502/python-poetry-how-to-install-optional-dependencies). Poetry will completely ignore them, but conda-lock will still treat them like dev dependencies and they will be optional. If the user specified `optional = false` they will be set as required in the lockfile.

### Conda Info

```shell
N/A
```

### Conda Config

```shell
N/A
```

### Conda list

```shell
N/A
```

## Suggested Fix for Situation 1

We can do one of the following:
1) Include optional dependencies, but in `conda-lock`, mark the category as main and required (assuming in general that if the category is main, it is a required dependency
Should we throw a warning indicating that the optional flag in the `pyproject.toml` was ignored?
2) Skip optional dependencies (with or without warning)
3) Throw an error if invalid optional dependencies are found
4) Create a new category for these

I would say lets go with option (1) and throw a warning.

## Suggested Fix for Situation 2

We can do one of the following:
- Include optional dependencies, but in `conda-lock`, mark the category as dev and optional. Warning or no warning?
- Skip optional dependencies (with or without warning)
- Throw an error if invalid optional dependencies are found
- Create a new category for these

Again, I would say lets go with option (1) and throw a warning.

srilman:
Fixes #385. This is a compilation of some unusual behavior I noticed when `conda-lock` parses some dependencies in `pyproject.toml` files with conflicting attributes:
- When a dependency is an extra, but not specified as optional
- When a dependency is an extra, but not in [tool.poetry.dependencies]
- When a dependency is specified as optional, but not in [tool.poetry.dependencies]
- When a dependency is specified as optional and in [tool.poetry.dependencies], but is not part of any extra 

There are all things that Poetry seems to handle quietly in its own way (see the related issue). To minimize the number of parsing changes, I did not change how `conda-lock` treats the resulting categories of dependencies. 
- If a dependency is part of an extra, but not specified as optional or not under [tool.poetry.dependencies], `conda-lock` will print a warning but still treat it as part of the extra as it always has.
- If a dependency is marked as optional or not optional but isn't associated with an extra, `conda-lock` will print a warning but just presume that it is optional if not in the main category, required if in main.

Essentially, this PR ensures that we maintain the invariant that an optional source dependency is a dependency not part of the main category. This is important for future PRs implementing multiple category support.


maresb:
I'm really looking forward to this PR series, but unfortunately I have no time this week.

srilman:
No worries @maresb. Always appreciate your reviews!

PR #389, opened 2023-03-13
Remove `optional` From Dependency Info in Internal Code

srilman:
This PR removes any references to an `optional` flag or option from the entire codebase. This builds on top of PR #388, so make sure to get that one merged first. At that point, there is an implicit invariant for all dependencies; the dependency is optional if `category != "main"`. Therefore, storing the optional value just makes the code unnecessarily complex.

Note that this does not affect how source files are parsed or the produced lock file in any way. The step that writes the lockfile out will insert `optional: true` or `optional: false` flags depending on the category.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -542,7 +542,7 @@ def render_lockfile_for_platform(  # noqa: C901
     lockfile.toposort_inplace()
 
     for p in lockfile.package:
-        if p.platform == platform and ((not p.optional) or (p.category in categories)):
+        if p.platform == platform and p.category in categories:
             if p.manager == "pip":
                 pip_deps.append(p)
             elif p.manager == "conda":
--- a/conda_lock/lockfile/__init__.py
+++ b/conda_lock/lockfile/__init__.py
@@ -21,7 +21,7 @@ from .models import TimeMeta as TimeMeta
 from .models import UpdateSpecification as UpdateSpecification
 
 
-def _apply_categories(
+def apply_categories(
     requested: Dict[str, Dependency],
     planned: Dict[str, LockedDependency],
     categories: Sequence[str] = ("main", "dev"),
--- a/conda_lock/lockfile/__init__.py
+++ b/conda_lock/lockfile/__init__.py
@@ -81,7 +81,6 @@ def _apply_categories(
         # try a conda target first
         target = seperator_munge_get(planned, dep)
         target.category = source.category
-        target.optional = source.optional
 
 
 def parse_conda_lock_file(
--- a/conda_lock/lockfile/__init__.py
+++ b/conda_lock/lockfile/__init__.py
@@ -3,7 +3,7 @@ import pathlib
 
 from collections import defaultdict
 from textwrap import dedent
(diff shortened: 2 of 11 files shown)


conda_lock/conda_lock.py
```diff
@@ -542,7 +542,7 @@ def render_lockfile_for_platform(  # noqa: C901
     lockfile.toposort_inplace()
 
     for p in lockfile.package:
-        if p.platform == platform and ((not p.optional) or (p.category in categories)):
+        if p.platform == platform and p.category in categories:
```
maresb, line 545: Note to self: `not p.optional` ⇒ `p.category == "main"`.
Since `"main" in categories`, this is redundant and can be deleted.

Therefore, this is logically sound.

conda_lock/lockfile/__init__.py
```diff
@@ -95,6 +93,9 @@ def parse_conda_lock_file(
     if not (isinstance(version, int) and version <= Lockfile.version):
         raise ValueError(f"{path} has unknown version {version}")
 
+    for p in content["package"]:
+        del p["optional"]
```
maresb, line 97: Note to self: we need to delete the `optional` key in order to parse this as a `StrictModel`.

maresb (approved):
From a thorough reading of the changes, I'm quite confident that this is a pure refactor. LGTM, and significantly increases code clarity.

If you could rebase on main I'd like to merge. Thanks @srilman!

maresb:
Or does it make more sense to merge #390 all at once, once I've completed that?

maresb:
Ah, there are serious merge conflicts with #290 now in `main`. :disappointed:

It looks like you're both doing similar things, so I wonder if this also fixed the same bugs.

srilman:
@maresb Resolving conflicts now ...

srilman:
Merged with main, everything should be good!

maresb:
Looking great, thanks!!!

As we wait for the Windows test to finish, can we rebase #390 on this?

conda_lock/lockfile/__init__.py
```diff
@@ -207,13 +207,22 @@ def write_section(text: str) -> None:
...
-                    content.json(by_alias=True, exclude_unset=True, exclude_none=True)
-                ),
-            },
-            stream=f,
-            sort_keys=False,
-        )
+        output: Dict[str, Any] = {
+            "version": Lockfile.version,
+            "metadata": json.loads(
+                content.metadata.json(
```
maresb, line 213: I just noticed that in #390 you have here instead `"metadata": content.metadata.dict(` which seems more elegant. Is this on purpose, or does this cover some edge case?

maresb:
The tests are passing. Let's pick up on https://github.com/conda/conda-lock/pull/389#discussion_r1199661413 with #390. I have to stop for tonight.

conda_lock/lockfile/__init__.py
srilman: I believe there was an issue when encoding `content.metadata` as a dict due to a special class / container that was used in the metadata (was it a frozenset?). Let me check for certain.

Issue #392, opened 2023-03-16
Github pip dependencies: InvalidRequirement: Expected end or semicolon


mjlbach:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Since https://github.com/conda/conda-lock/issues/4 is closed (and a feature request), I'm reopening this as a bug report:

1. Make the following  `environment.yml`

```yaml
name: test                                                                                                                                                                      
channels:                                                                               
  - conda-forge                                                                                                                                                                 
dependencies:                                                                           
  - python=3.10                             
  - pip:                                                                                
    - xarray                                                                                                                                                                    
    - git+https://github.com/pandas-dev/pandas.git@v1.4.4     
```

2. run `conda-lock -f environment.yml` and see the following error:
```
  File "/home/michael/.condax/conda-lock/lib/python3.11/site-packages/pkg_resources/__init__.py", line 3211, in parse
    (req,) = parse_requirements(s)
    ^^^^^^
  File "/home/michael/.condax/conda-lock/lib/python3.11/site-packages/pkg_resources/__init__.py", line 3170, in __init__
    super(Requirement, self).__init__(requirement_string)
  File "/home/michael/.condax/conda-lock/lib/python3.11/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 37, in __init__
    raise InvalidRequirement(str(e)) from e
pkg_resources.extern.packaging.requirements.InvalidRequirement: Expected end or semicolon (after name and no valid version specifier)
    git+https://github.com/pandas-dev/pandas.git@v1.4.4
```

Seems like the issue is passing 
```
(Pdb++) requirement_specifier
'git+https://github.com/pandas-dev/pandas.git@v1.4.4'
```
to
`weparsed_req = Requirement.parse(requirement_specifier)`

I'm not 100% sure how to solve this, I think everything should move to using the new  `from packaging.requirements import Requirement` interface as `from pkg_resources import Requirement` is deprecated, but that doesn't solve the parsing issue (hacked up a WIP PR but it's non functional)

Note adding a name to the requirements solves this particular issue but 

```yaml
name: test                                                                                                                                                                      
channels:                                                                               
  - conda-forge                                                                                                                                                                 
dependencies:                                                                           
  - python=3.10                             
  - pip:                                                                                
    - xarray                                                                                                                                                                    
    - pandas @ git+https://github.com/pandas-dev/pandas.git@v1.4.4     
```

[...]

PR #394, opened 2023-03-25
feat: add support for git to lockfile

mjlbach:
Closes https://github.com/conda/conda-lock/issues/392

This is an incredibly hacky solution for https://github.com/conda/conda-lock/issues/392, could use some additional direciton on making this more robust (does anyone know a package parser/upstream efforts to get package parsers that resolve git forges?)

### DIFF
--- a/conda_lock/models/lock_spec.py
+++ b/conda_lock/models/lock_spec.py
@@ -31,8 +31,11 @@ class URLDependency(_BaseDependency):
     url: str
     hashes: List[str]
 
+class VCSDependency(_BaseDependency):
+    source: str
+    vcs: str
 
-Dependency = Union[VersionedDependency, URLDependency]
+Dependency = Union[VersionedDependency, URLDependency, VCSDependency]
 
 
 class Package(StrictModel):
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -16,6 +16,7 @@ from conda_lock._vendor.poetry.core.packages import (
     ProjectPackage as PoetryProjectPackage,
 )
 from conda_lock._vendor.poetry.core.packages import URLDependency as PoetryURLDependency
+from conda_lock._vendor.poetry.core.packages import VCSDependency as PoetryVCSDependency
 from conda_lock._vendor.poetry.factory import Factory
 from conda_lock._vendor.poetry.installation.chooser import Chooser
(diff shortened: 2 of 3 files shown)


maresb:
Awesome! Looks like a great start. Thanks so much! I hope to take a closer look soon.

mjlbach:
Thanks, ended up being less code than I expected. I wish their was a more robust package for parsing the vcs + url + hash, I thought that pkg_resource being deprecated (https://setuptools.pypa.io/en/latest/pkg_resources.html) the replacement (packaging.requirements import Requirement](https://packaging.pypa.io/en/stable/requirements.html#requirements) would have some better utilities for this...

maresb:
Yes, we should migrate that. Feel free to be bold with your PR. ;)

mjlbach:
Just leaving this here, here is how poetry core deals with the vcs problem (similar) 

https://github.com/python-poetry/poetry-core/blob/e33e9d1613338f045b55cae1a7953dec7d86f311/src/poetry/core/packages/dependency.py#L414-L427

mjlbach:
@maresb in the process of being bold, I think we can re-use some upstream poetry handling around Requirements/link parsing, although if you want to in the long run remove the vendored dependencies in favor of a custom solver that might make de-integration slightly more complicated.

Would it be possible to get a poetry bump merged soon https://github.com/conda/conda-lock/pull/310?

maresb:
Leaving the door to Poetry de-integration is really important to me. That said, I would really like to bump the vendored Poetry. Also, as opposed to importing directly from `_vendored`, I'd like to set up an intermediate `interfaces` module where we can see at-a-glance which vendored stuff we're actually using.

Doing the vendoring requires me finding a good chunk of uninterrupted time where I can see through the whole process. Over the next month or so I have a lot of major drains on my personal time, so I can't really promise anything, but I'll try to get to it soon.

mjlbach:
Yeah 100%, totally understand the time constraints situation. I'm just excited to see this feature in conda-lock and want to do what I can (coding-wise) to make that happen :) Just lmk where I should go from here, for now this fork lets me lock chumpy from git haha.

muyajil:
Is there motivation to continue this? I think it would be good to support the `git+ssh` and `git+https` prefixes such that we can lock complete environments.
Is there sth that can be done to help progress this?

maresb:
Already merged in #435

PR #398, opened 2023-04-12
Update the `seperator_munge_get` to provide flexibility for dependency lookup

leeleavitt:
The fix in this PR adresses two example environment yamls,

In this example `bullet-python` a dependency of `pytest-wdl` breaks `seperator_munge_get`
```
name: python
platforms:
  - linux-64
  - osx-64
channels:
  - bioconda
  - conda-forge
dependencies:
  - python=3.9.12
  - pip
  - miniwdl=1.2.3
  - pip:
    - pytest-wdl
```

In this example `python-tzdata` breaks `seperator_munge_get`,

```
name: python
platforms:
  - linux-64
  - osx-64
channels:
  - bioconda
  - conda-forge
dependencies:
  - python=3.9.12
  - pybedtools
  - pip
  - pip:
    - pandas==1.3.5
```

### DIFF
--- a/conda_lock/lockfile/__init__.py
+++ b/conda_lock/lockfile/__init__.py
@@ -41,7 +41,19 @@ def _apply_categories(
             try:
                 return d[key.replace("-", "_")]
             except KeyError:
-                return d[key.replace("_", "-")]
+                try:
+                    return d[key.replace("_", "-")]
+                except KeyError:
+                    # for bullet-python or python-tzdata
+                    try:
+                        attempted_split = key.split("_")[0].split("-")[0]
+                        if attempted_split == 'python':
+                            attempted_split = key.split("_")[-1].split('-')[-1]
+                        return d[attempted_split]
+                    except KeyError:
+                        list(d.keys())
+                        raise KeyError(f"Could not find {key} in list of dependencies: {list(d.keys())}")
+
 
     for name, request in requested.items():
         todo: List[str] = list()


maresb:
Thank you very much for this @leeleavitt!!!

I have been meaning to dig into this function to try to understand exactly what it's supposed to accomplish. There are some very helpful comments in #290 about what the surrounding code is doing.

I would like to merge #290 very soon. Would you be able to test your update with that branch?

Also, since you have spent some time on `separator_munge_get`, do you understand exactly what it is doing? If so, would you be able to explain it? In particular, I am wondering why for pip dependencies we aren't using [canonicalize_name](https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_name), and I am also wondering what exactly are the ["inconsistencies" within Conda](https://github.com/conda/conda-lock/pull/398/files#diff-481bd517187731b0ecd0cfa54dc36de501405dcca83885b9253bb7fe50287f3cL37).

Cakell:
Hi @maresb .

Regarding the logic behind `separator_munge_get`, see #139 and in particular [this commit](https://github.com/conda/conda-lock/pull/139/commits/05028121dc1899065b53fbf14ebb291b6c59279f).

TL;DR:

This code is meant to "Add some slightly more robust fallback handling to find packages that have different separators when using a mixed poetry+conda solve".

Hopes this helps.

PS. I also came across this `separator_munge_get` bug, and would love to see a release which fixes this soon : )

maresb:
Thanks @Cakell, does this PR solve the bug for your case?

olivier-lacroix:
> Thanks @Cakell, does this PR solve the bug for your case?

It does for me! encountered the issue with python-tzdata

leeleavitt:
I will see if i can refactor the code to get around the flake8 complexity failure. Is there a specific version of flake8 or black to use? I couldn't see a version pinned in the repo. Should i just go with the latest versions?

Also @maresb in regards to testing #290 should I merge that branch into mine and then test it?

maresb:
Thanks @leeleavitt!

To get around the complexity failure, I think you could simply do

```python
try:
    ...
except KeyError:
   pass
```

and this should allow you to dedent the rest, and I expect this to decrease your complexity score.

> should I merge that branch into mine and then test it?

Exactly. Thanks so much for your work on this, I really appreciate it!

leeleavitt:
> Thanks @leeleavitt!
> 
> To get around the complexity failure, I think you could simply do
> 
> ```python
> try:
>     ...
> except KeyError:
>    pass
> ```
> 
> and this should allow you to dedent the rest, and I expect this to decrease your complexity score.
> 
> > should I merge that branch into mine and then test it?
> 
> Exactly. Thanks so much for your work on this, I really appreciate it!

Alright i spent some time testing this yesterday and #290 fixes the issue without the need for my fix. I'm going to write up my findings on that PR.

lesteve:
Actually I did not see this one before opening #401 but it looks like both are related. In #401 case it is not fixed by #290.

Removing `python-` does not seem like super robust. I think there should be a way somewhere to say something like:
- conda `tzdata` is not the same as pip `tzdata`
- conda `python-tzdata` is the same as pip `tzdata`, this is already done I am guessing by the conda pip mapping

PR #406, opened 2023-05-18
Fix import in _vendor

epassaro:
### Description

Found an incorrect import that breaks `conda-lock` when `toolz` is not installed.

### DIFF
--- a/conda_lock/_vendor/conda/models/version.py
+++ b/conda_lock/_vendor/conda/models/version.py
@@ -10,7 +10,7 @@ from itertools import zip_longest
 try:
     from tlz.functoolz import excepts
 except ImportError:
-    from conda_lock._vendor.conda._vendor.toolz.functoolz import excepts
+    from conda_lock._vendor.conda._vendor.toolz import excepts
 
 from ..exceptions import InvalidVersionSpec
 


maresb:
Thanks, but I'm a bit confused about the context. Toolz is a dependency, so how would someone even get into the situation of running conda-lock without toolz being installed?

epassaro:
> Thanks, but I'm a bit confused about the context. Toolz is a dependency, so how would someone even get into the situation of running conda-lock without toolz being installed?

I was doing a weird experiment, trying to "pack" `conda-lock` with `pyinstaller`, and step up with this issue. I abandoned the idea quickly, but I thought this should be fixed anyways.

You can check this behavior by force removing `toolz` from the environment.

maresb:
Looks like this was already done upstream in https://github.com/conda/conda/pull/12016, so next time we revendor conda, that will get pulled in.

Unless there's a really compelling reason I'd opt against merging this, because the point of vendoring is that you delegate maintenance to upstream, and this would bring us out-of-sync with the vendored version. A weird experiment isn't so compelling.

By the way, taking a wild and likely incorrect guess about your motivation for trying pyinstaller, if it was to decouple conda-lock from your active environment due to dependency conflicts, then I'd recommend installing conda-lock via condax or pipx.

epassaro:
Thank you! @maresb

Issue #408, opened 2023-05-18
Simple reproducer for virtual PackageRecord error


maresb:
This is a lingering point of major breakage. Fails on both 1.4.0 and `main`.

Seems related to #202, #292, #381.

```bash
docker run --rm -it mambaorg/micromamba
micromamba install -y -c conda-forge mamba conda-lock git
mamba --version
# mamba 1.4.2
# conda 23.3.1
cat <<EOF > environment.yml
channels:
- conda-forge
dependencies:
- git
platforms:
- osx-64
EOF
conda-lock
pip install git+https://github.com/conda/conda-lock@4a667bd  # main, right after merging #290
conda-lock
```

<details>
<summary>STDERR:</summary>

```python
PackageRecord(_hash=-2262511397987521167, name='__osx', version='0', build='0', build_number=0, channel=Channel("@"), subdir='osx-64', fn='__osx', md5='12345678901234567890123456789012', package_type='virtual_system')

Traceback (most recent call last):
  File "/opt/conda/bin/conda-lock", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #409, opened 2023-05-18
Add test for GH #408 PackageRecord error

maresb:
I'm not sure what's causing #408 but I want to write a test so we can check when it's solved.

### DIFF
--- a/tests/test-git-package/environment.yml
+++ b/tests/test-git-package/environment.yml
@@ -1,0 +1,6 @@
+channels:
+- conda-forge
+dependencies:
+- git
+platforms:
+- osx-64
--- a/tests/test_conda_lock.py
+++ b/tests/test_conda_lock.py
@@ -1382,6 +1382,20 @@ def test_run_lock_relative_source_path(
     assert [p.resolve() for p in src_files] == [environment.resolve()]
 
 
+@pytest.fixture
+def test_git_package_environment(tmp_path: Path):
+    return clone_test_dir("test-git-package", tmp_path).joinpath("environment.yml")
+
+
+def test_git_gh_408(
+    monkeypatch: pytest.MonkeyPatch, test_git_package_environment: Path, conda_exe: str
+):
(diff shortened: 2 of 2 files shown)


ogrisel:
I recently observed the `KeyError` for the `__osx` package when running a conda-lock with recent mamba & conda installed on my system.

I contributed a fix to the upstream repo that works for me:

- https://github.com/mamba-org/mamba/pull/2923

I am not sure if this is related to this or not.

/cc @lesteve.

maresb:
Ah, bad news, but thanks for letting us know. Also thanks for submitting the upstream fix! And also for raising visibility on this PR which I meant to merge.

maresb:
@ogrisel, can you share your `environment.yml` file which led to the error?

ogrisel:
```
conda-lock lock --mamba --kind explicit --platform osx-64 --file build_tools/azure/pylatest_conda_forge_mkl_osx-64_environment.yml --filename-template build_tools/azure/pylatest_conda_forge_mkl_osx-64_conda.lock
```

and here is the content of `pylatest_conda_forge_mkl_osx-64_environment.yml`:

```yaml
channels:
  - conda-forge
dependencies:
  - python
  - numpy
  - blas[build=mkl]
  - scipy
  - cython
  - joblib
  - threadpoolctl
  - matplotlib
  - pandas
  - pyamg
  - pytest
  - pytest-xdist=2.5.0
  - pillow
  - setuptools
  - pytest-cov
  - coverage
  - ccache
  - compilers
  - llvm-openmp
```

lesteve:
For the record this command runs fine for me on a Linux machine, so maybe the issue only happens when running the command on an OSX machine?

ogrisel:
Yes I believe that this might be because the `__osx` virtual package does not show up on a Linux host.

This is probably a bug of mamba or conda that ignores the `--platform osx-64` argument when defining the system specific virtual packages.

Issue #411, opened 2023-05-20
No changes to LockedDependency allowed without schema change


maresb:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Merging #389, in particular 08f2292, broke `conda-lock install` because `LockedDependency` is a `StrictModel`, so removing `optional` means existing lockfiles which contain `optional` now have fields which are not permitted and lead to an error. CC @srilman

PR #412, opened 2023-05-21
Create a preliminary internal v2 lockfile schema but enforce v1

maresb:
As described in #411, the merge of #389 broke parsing of lockfiles for the `conda-lock install` command due to a schema change.

In this PR I create explicit v1 and v2 schema, the v2 being preliminary. The only difference between the two is that v2 omits the `optional` from packages. There is no loss of information thanks to the invariant `optional = category != "main"`. Thus it's easy to convert between v1 and v2, and the corresponding functions are added here.

The combination of #389 and this PR should be a *pure refactor*, as input and output are done exclusively with the v1 schema.

This sets us up to make the schema modifiable, as there are some changes I'd like to make. (RFC coming soon...)

Before I consider this to be closing for #411, I want to add tests for `conda-lock install`. I merged #411 because the tests were green, despite the breakage, so I'd like to avoid this happening in the future.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -54,7 +54,7 @@ from conda_lock.invoke_conda import (
     is_micromamba,
 )
 from conda_lock.lockfile import parse_conda_lock_file, write_conda_lock_file
-from conda_lock.lockfile.models import (
+from conda_lock.lockfile.v1.models import (
     GitMeta,
     InputMeta,
     LockedDependency,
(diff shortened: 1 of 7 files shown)


maresb:
@srilman, I'd appreciate a review on this if you get the chance. Also, I'm going to try to rebase #390 on this soon.

maresb:
I'm trying to update my #410 which rebases #390 so that the previous logic is reproduced. It it much trickier than I hoped. It's more of a logic puzzle than I can solve for tonight.

My goal would be to adjust #390 so that `categories` is not a `set` but a `list`. And that `list` should have the property that its first (or last) element should be the `category` selected previously.

This way, lockfile conversion would work as follows: 
* v1 → v2, `categories = [category]`
* v2 → v1, `category = categories[0]`

maresb:
I don't like having a broken `main`, so I'm going to merge this. We can revert as needed.

Issue #414, opened 2023-05-24
Error processing dependency when writing lockfile


AlbertDeFusco:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

locking following environment.yml file fails

```yaml
channels: [defaults]
dependencies: [cryptography]
platforms: [linux-64]
```

Here's the error message

```
❯ conda-lock lock
Locking dependencies for ['linux-64']...
INFO:conda_lock.conda_solver:linux-64 using specs ['cryptography']
Traceback (most recent call last):
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/conda_lock/src_parser/__init__.py", line 488, in seperator_munge_get
    return d[key]
KeyError: 'openssl<1.1.2a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/conda_lock/src_parser/__init__.py", line 491, in seperator_munge_get
    return d[key.replace("-", "_")]
KeyError: 'openssl<1.1.2a'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/adefusco/Applications/miniconda3/bin/conda-lock", line 11, in <module>
    sys.exit(main())
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/Users/adefusco/Applications/miniconda3/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #415, opened 2023-05-24
use matchspec in LockedDependency creation (issue 414)

AlbertDeFusco:
This addresses #414 

### Description

The previous implementation assumed that dependencies of installed packages listed in the dry-run-solve json output are of the format "\<name\> \<version\>", which is a subset of MatchSpec.


maresb:
Thank you very much for this, @AlbertDeFusco. It's indeed much better to replace such string manipulations with proper classes.

PR #418, opened 2023-05-26
support nodefaults in environment.yml specs

### LINKED ISSUE #192: Cannot lock when Conda channel_alias is set and nodefaults channel is in environment.yml
Having a `.condarc` like this:

```yaml
channel_alias: example.org
```

And an `environment.yml` likes this:

```yaml
channels:
  - conda-forge
  - nodefaults
packages:
  - ...
```

Causes the following error when trying to lock the environment:

```
Generating lockfile(s) for linux-64...
Could not lock the environment for platform linux-64
The channel is not accessible or is invalid.
  channel name: nodefaults
  channel url: example.org/nodefaults
  error code: 404
```

You can resolve the issue by adding the following to your `.condarc`:

```yaml
custom_channels:
  nodefaults: https://conda.anaconda.org
```

However, I wouldn't expect this to be necessary, as `nodefaults` is not an actual channel. It is only supposed to indicate that no default channels should be used.

I've tried this on conda-lock 0.13.2 and 1.0.5. My conda/mamba version is:

```
mamba 0.23.0
conda 4.12.0
```

It is important to mention that you can `mamba env create` the environment. So the issue seems to lie within conda-lock.

Any idea what might be causing this?

FHTMitchell:
### Description

conda environment.yml files [support](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually) a `- nodefaults` "channel" which suppresses the use of defaults channels in creating the environment. `conda-lock` uses `--override-channels` in the env creation anyway so `- nodefaults` is redundant but it causes a 404 error in the build since we treat `nodefaults` as a real channel. 

This fixes #192

### DIFF
--- a/conda_lock/src_parser/environment_yaml.py
+++ b/conda_lock/src_parser/environment_yaml.py
@@ -114,6 +114,11 @@ def parse_environment_file(
 
     env_yaml_data = yaml.safe_load(content)
     channels: List[str] = env_yaml_data.get("channels", [])
+    try:
+        # conda-lock will use `--override-channels` so nodefaults is redundant.
+        channels.remove("nodefaults")
+    except IndexError:
+        pass
 
     # These extension fields are nonstandard
     category: str = env_yaml_data.get("category") or "main"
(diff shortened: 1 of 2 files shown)


maresb (approved):

maresb:
Thanks @FHTMitchell!!!

PR #425, opened 2023-06-01
Allow the use of channels like `channel::package` in TOML files.

romain-intel:
This functionality worked for environment.yml type of files but was broken for TOML files.

### Description

Small change to allow the use of `channel::package` in the `tool.conda-lock.dependencies` section of TOML files.

A test is included. The test fails before this PR and succeeds after. The test specifically includes a package that is available on the channel specified but *not* on the default conda-forge channel.

### DIFF
--- a/conda_lock/src_parser/pyproject_toml.py
+++ b/conda_lock/src_parser/pyproject_toml.py
@@ -279,13 +279,22 @@ def specification_with_dependencies(
     ).items():
         if isinstance(depattrs, str):
             conda_version = depattrs
+            channel_split = depname.split("::")
+            if len(channel_split) == 1:
+                channel = None
+                name = channel_split[0]
+            else:
+                channel = channel_split[0]
+                name = channel_split[1]
+
             dependencies.append(
                 VersionedDependency(
-                    name=depname,
+                    name=name,
                     version=conda_version,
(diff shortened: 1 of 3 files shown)


maresb:
Thanks a lot for this @romain-intel!

It looks like we're starting to manipulate structured data as strings, which tends to eventually lead to issues. I wonder if it would instead make more sense here to lean on [conda_spec_to_versioned_dep](https://github.com/conda/conda-lock/blob/main/conda_lock/src_parser/conda_common.py).

I personally don't understand the design decision of making `[tool.conda-lock.dependencies]` a dict instead of a list. (It feels like Poetry-motivated insanity.) Consequently, instead of a nice `MatchSpec` string we now have this awkward `depname` `depattrs` key-value-pair. Optimistically, perhaps `conda_spec_to_versioned_dep(f"{depname} {depattrs}", "main")` might work?

romain-intel:
That's probably a better solution and more in line with what the environment.yaml does. I can make that change and try it out.

romain-intel:
There we go -- looks even better. Tiny change :)

maresb:
Woah, I can't believe that actually worked! :joy: Nicely done!

maresb (approved):

PR #427, opened 2023-06-04
Change to temp directory for running tests

maresb:
Conda-lock is sensitive to files in the current directory. Using it from the project root risks conflicts, like in #359 where I'd like to add a `conda-lock.yml`. Thus it may be desirable to run from an isolated directory by default.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import re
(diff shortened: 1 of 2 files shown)


maresb:
I don't want to merge this anymore. I'm satisfied now that the tests are no longer polluting the current directory.

Issue #426, opened 2023-06-02
Remove cuda from default virtual packages?


minrk:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

I noticed while installing pytorch-cpu that it pulled in the cuda variant of libhwloc and thereby cudatoolkit, doubling the size of my image. I tracked it down to the default virtual package spec assuming all machines are likely to have cuda by default.

I was able to solve it with a custom virtual packages spec (capturing the virtual packages from `conda info` in the base image), but it seems to make more sense to me for cuda to be opt-in instead of opt-out, since it's unvailable more often than not, and the cost of an incorrect assumption of its presence is high (massive size increase, non-working packages), while the cost of missing it is low (reduced performance or informative, immediate error if cuda is actually required to install). Or is there a consideration I'm missing?

### Conda Info

```shell
active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /home/user/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 23.3.1
    conda-build version : not installed
         python version : 3.10.11.final.0
       virtual packages : __archspec=1=x86_64
                          __glibc=2.35=0
                          __linux=5.15.0=0
                          __unix=0=0
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/minrk/linux-64
                          https://conda.anaconda.org/minrk/noarch
          package cache : /opt/conda/pkgs
                          /home/user/.conda/pkgs
       envs directories : /opt/conda/envs
                          /home/user/.conda/envs
               platform : linux-64
             user-agent : conda/23.3.1 requests/2.31.0 CPython/3.10.11 Linux/5.15.0-1031-gcp ubuntu/22.04.2 glibc/2.35
                UID:GID : 1000:100
             netrc file : None
           offline mode : False
```

### Conda Config

```shell
==> /opt/conda/.condarc <==
auto_update_conda: False
channels:
  - conda-forge
show_channel_urls: True
```

### Conda list

```shell
# packages in environment at /Users/minrk/conda:
#
# Name                    Version                   Build  Channel
aiofiles                  22.1.0             pyhd8ed1ab_0    conda-forge
[...]

PR #429, opened 2023-06-08
accept build strings in virtual package specs

minrk:
FakePackage didn't accept build strings, so one couldn't have e.g. `__archspec=1=x86_64`

Found while exploring #426


mariusvniekerk (approved):

maresb:
Thanks @minrk!

PR #437, opened 2023-06-15
List prefixes for consideration for env var substitution

nicoddemus:
This explicitly lists the prefixes used by conda-lock to detect and use environment variables to replace sensitive variables in lock files, taken from here:

https://github.com/conda/conda-lock/blob/60991f485f5711462b574a290ce7000e40aa0b9d/conda_lock/models/channel.py#L198-L226C20

### DIFF
--- a/docs/authenticated_channels.md
+++ b/docs/authenticated_channels.md
@@ -112,7 +112,7 @@ package:
   version: 22.02.00
 ```
 
-Note however that the rendered lockfiles (` --kind explicit`) will contain substituted environment variables, so if you are making use of `conda-lock`
+Note however that the rendered lockfiles (`--kind explicit`) will contain substituted environment variables, so if you are making use of `conda-lock`
 in conjunction with git these should **NOT** be checked into version control.
 
 [anaconda.org]: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually
--- a/docs/authenticated_channels.md
+++ b/docs/authenticated_channels.md
@@ -78,7 +78,7 @@ Since we can generally assume that these substitutions are both volatile _and_ s
 the raw version of a url in the unified lockfile.
 
 If it encounters a channel url that looks as if it contains a credential portion (see below) it will search the currently
-available environment variables for a match with that variable. 
+available environment variables for a match with that variable.
 
 `conda-lock` will identify the following environment variables as containing credentials only if they have these *suffixes*:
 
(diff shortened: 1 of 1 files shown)


docs/authenticated_channels.md
```diff
@@ -77,25 +77,42 @@ Additionally simple auth also support the [--strip-auth, --auth and --auth-file]
...
+If it encounters a channel url that looks as if it contains a credential portion (see below) it will search the currently
+available environment variables for a match with that variable. 
+
+`conda-lock` will identify the following environment variables as containing credentials only if they have these *suffixes*:
+
+* User names: `["USERNAME", "USER"]`.
+* Passwords: `["PASSWORD", "PASS", "TOKEN", "KEY"]`.
+* Tokens: `["TOKEN", "CRED", "PASSWORD", "PASS", "KEY"]`.
+
+In the case of a match that portion of the url will be replaced with a environment variable.
```
nicoddemus, line 89:
```suggestion
In the case of a match that portion of the url will be replaced with an environment variable.
```

docs/authenticated_channels.md
```diff
@@ -77,25 +77,42 @@ Additionally simple auth also support the [--strip-auth, --auth and --auth-file]
...
+    - QUETZ_API_KEY
+  - url: conda-forge
+    used_env_vars: []
+package:
+- platform: linux-64
+  url: https://host.tld/t/$QUETZ_API_KEY/channel_name/linux-64/libsomethingprivate-22.02.00.tar.bz2
+  version: 22.02.00
+```
 
 The rendered lockfiles will contain substituted environment variables, so if you are making use of `conda-lock`
```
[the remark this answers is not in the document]
mariusvniekerk: The lockfiles in the conda-explicit form will contain these substituted in as conda does not support environment variables

mariusvniekerk (approved):

docs/authenticated_channels.md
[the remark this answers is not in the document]
nicoddemus: Ahh got it, will clarify the statement.

docs/authenticated_channels.md
```diff
@@ -77,25 +77,42 @@ Additionally simple auth also support the [--strip-auth, --auth and --auth-file]
...
```
nicoddemus, line 115:
```suggestion
Note however that the rendered lockfiles (` --kind explicit`) will contain substituted environment variables, so if you are making use of `conda-lock`
```

docs/authenticated_channels.md
```diff
@@ -77,25 +77,42 @@ Additionally simple auth also support the [--strip-auth, --auth and --auth-file]
...
 in conjunction with git these should _NOT_ be checked into version control.
```
nicoddemus, line 116:
```suggestion
Note however that the rendered lockfiles (` --kind explicit`) will contain substituted environment variables, so if you are making use of `conda-lock`
in conjunction with git these should **NOT** be checked into version control.
```

nicoddemus:
(This should be squashed, let me know if you prefer I do it from the command-line rather during merging).

PR #439, opened 2023-06-23
Create macOS platform tags using packaging 

jjhelmus:
### Description

Populate the list of supported platforms tags for the macOS platform using the `mac_platforms` function in packaging.tags.  This given a complete listing of platform tags which is also used by pip.  

This avoiding the need to periodically update a static listing of tags.  The new `MACOS_VERSION` variable will need to be updated periodically to reflect the latest version of macOS.

Likely fixes #403 and #436

PR #444, opened 2023-06-28
Move other regression tests to test_regression

mariusvniekerk:
As part the work on #439 I introduced a new test file for the regressions in order to keep the tests a little cleaner.

This moves the other explicit regression tests over to the new location

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,13 +1,21 @@
 [mypy]
 plugins = pydantic.mypy
-disallow_untyped_defs = true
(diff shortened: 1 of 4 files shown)


maresb (approved):
Makes the mypy config more explicit and uniform over tests/.

Moves around a few fixtures and tests to different modules.

LGTM

maresb:
I've been wanting to move the gdal test within pytest. That should be easy now with that new fixture, right?

PR #445, opened 2023-06-30
from-env subcommand for creating lock files from existing environments

jjhelmus:
This is a rough version of a new `from-env` subcommand that creates a lock file from an existing conda environment. 

I wanted to submit the change in this form to see if there was interest in such a feature and to collected feedback on the approach. If this is something that would be accepted into conda-lock, I'd be happy to continue working on the change to get it into a more complete state.


jjhelmus:
An example of the new subcommand:
```
❯ conda create --prefix ./example python=3.10 numpy
...
❯ conda-lock from-env --prefix ./sample 
 - Install lock using: conda-lock install --name YOURENV conda-lock.yml
❯ head -40 conda-lock.yml
#
#
#     conda-lock lock  --lockfile conda-lock.yml --update PACKAGE
#     conda-lock  --lockfile conda-lock.yml
version: 1
metadata:
  content_hash: {}
  channels:
  - url: defaults
    used_env_vars: []
  platforms:
  - osx-arm64
  sources: []
package:
- name: ca-certificates
  version: 2023.05.30
  manager: conda
  platform: osx-arm64
  dependencies: {}
  url: https://repo.anaconda.com/pkgs/main/osx-arm64/ca-certificates-2023.05.30-hca03da5_0.conda
  hash:
    md5: 323db96c02c0e6ee266388ad4be3f762
    sha256: 98ad45e8ca832f8b9ee0faabc8791ff9d6488d9d2d0b5106038f7bad6d712436
  category: main
  optional: false
- name: libcxx
  version: 14.0.6
  manager: conda
  platform: osx-arm64
  dependencies: {}
  url: https://repo.anaconda.com/pkgs/main/osx-arm64/libcxx-14.0.6-h848a8c0_0.conda
  hash:
  ```

maresb:
Thanks a lot for this! Looks very interesting.

I'm very curious what is the motivation and what this accomplishes that `conda env export` doesn't.

AlbertDeFusco:
I like this approach over `conda env export` in that the architecture is stated in the lock file. If someone gives you the output of `conda env export` it's not immediately apparent what platform that was created from.

And if the `conda env export --from-history` output is retained along side the single-platform lockfile you could solve for other platforms while keeping the original platform lock fixed.

PR #448, opened 2023-07-05
Capture failed tests as actual step/job failures

riccardoporreca:
Should fix  #447

This is done by Enabling fail-fast bash behavior, see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

PR #451, opened 2023-07-07
Fix virtual package versions being detected as floats

maresb:
### Description

Looks like either Pydantic 2 parsing is breaking something or a failing test may have gotten merged.

This fixes it by quoting the `X.Y`-style version strings

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,11 +227,11 @@ If you want to override which virtual packages are injected you can create a fil
 subdirs:
   linux-64:
     packages:
-      __glibc: 2.17
-      __cuda: 11.4
+      __glibc: "2.17"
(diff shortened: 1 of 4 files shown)


riccardoporreca:
@maresb, I spotted also an instance of this in [docs/flags.md](https://github.com/maresb/conda-lock/blob/fix-virtual-package-floats/docs/flags.md#--virtual-package-spec)

Linking here also https://github.com/conda/conda-lock/pull/448#issuecomment-1622219058, which mentions the same issue

maresb:
Thanks a lot @riccardoporreca for the pointer, I really appreciate it. I had a long week, and I didn't really absorb your initial report.

PR #453, opened 2023-07-09
Include "fn" in fake_conda_environment for mamba>=1.4.6 compatibility

riccardoporreca:
### Description

See #452. The initial goal of this (draft) PR is to provide evidence about the fact that including the file name ("fn") in the fake environment packages' JSON file makes mamba>=1.4.6 happy

PR #455, opened 2023-07-09
Unpin micromamba (on #453)

maresb:
### Description

Same as #454 but rebased on #453


maresb:
Not necessary since #454 passed without #453.

PR #459, opened 2023-07-13
Update to pydantic 2.0

### LINKED ISSUE #472: conda-lock is incompatible with pydantic 2
### Checklist

### What happened?

I have conda-lock installed via `brew` (it uses pydantic 2 for the build) and when I enter `conda-lock render --help`, I get the following warnings:

```
❯ conda-lock render --help
/opt/homebrew/Cellar/conda-lock/2.1.1/libexec/lib/python3.11/site-packages/pydantic/_internal/_config.py:261: UserWarning: Valid config keys have changed in V2:
* 'json_encoders' has been removed
  warnings.warn(message, UserWarning)
/opt/homebrew/Cellar/conda-lock/2.1.1/libexec/lib/python3.11/site-packages/pydantic/_internal/_config.py:261: UserWarning: Valid config keys have changed in V2:
* 'allow_mutation' has been removed
  warnings.warn(message, UserWarning)
Usage: conda-lock render [OPTIONS] [LOCK_FILE]
[...]
```

### Conda Info

_No response_

### Conda Config

_No response_

### Conda list

_No response_

### Additional Context

_No response_

mariusvniekerk:
Closes #472

### DIFF
--- a/conda_lock/models/__init__.py
+++ b/conda_lock/models/__init__.py
@@ -1,11 +1,5 @@
 from pydantic import BaseModel
 
 
-class StrictModel(BaseModel):
+class StrictModel(BaseModel, extra="forbid"):
     """A Pydantic BaseModel forbidding extra fields and encoding frozensets as lists"""
-
-    class Config:
-        extra = "forbid"
-        json_encoders = {
-            frozenset: list,
-        }
--- a/conda_lock/models/__init__.py
+++ b/conda_lock/models/__init__.py
@@ -2,4 +2,4 @@ from pydantic import BaseModel
 
 
 class StrictModel(BaseModel, extra="forbid"):
-    """A Pydantic BaseModel forbidding extra fields and encoding frozensets as lists"""
+    """A Pydantic BaseModel forbidding extra fields"""
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     - id: check-ast
 
 - repo: https://github.com/psf/black
-  rev: 23.3.0
+  rev: 23.7.0
   hooks:
   - id: black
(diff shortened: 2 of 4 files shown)


maresb:
Are you sure we want to do this already?

The Pydantic v2 migration seems like a rough edge. I think this might make it a lot harder for people to install conda-lock in their base environments.

maresb:
Maybe we should instead suppress the warnings for a bit?

maresb:
In light of #462 I think this should be fine.

I'm just mildly concerned because I don't understand the effect of removing

```python
json_encoders = {frozenset: list}
```

maresb (approved):
I think this deserves its own release

maresb:
@mariusvniekerk, this is marked as draft, but I think it may be ready to merge?

mariusvniekerk:
Yeah just left it as draft so that we can land whatever we want before landing this breaking dependency

maresb:
@mariusvniekerk it looks like I forgot to provide this explanation:

The required new functionality exists in Pydantic 1.10, so pinning `>=2` isn't even necessary. I already added a commit in this branch to relax the pin to `>=1.10`. I tested the changes in #462 by pinning `pydantic <2`.

That should be fine, right?

conda_lock/models/__init__.py
```diff
@@ -1,11 +1,5 @@
 from pydantic import BaseModel
 
 
-class StrictModel(BaseModel):
+class StrictModel(BaseModel, extra="forbid"):
     """A Pydantic BaseModel forbidding extra fields and encoding frozensets as lists"""
```
mariusvniekerk, line 5:
```suggestion
    """A Pydantic BaseModel forbidding extra fields"""
```

Issue #461, opened 2023-07-15
Add support PyPi packages when using `--strip-auth` for `--kind lock` lockfiles


jacksmith15:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

https://github.com/conda/conda-lock/pull/323 added support for stripping credentials from PyPi packages in `--kind explicit` lockfiles.

I propose adding support for `--kind lock` lockfiles too, since these are the default.

### Why is this needed?

Currently running `conda-lock lock --strip-auth` when private PyPi repositories are configured results in credentials being written directly to the lockfile, meaning it cannot be committed to source control.

Currently this blocks users who have the following constraints

- Building for multiple platforms (unified lockfile)
- Depend on packages in private PyPi repositories (auth)

### What should happen?

Running `conda-lock lock --strip-auth --kind lock` should strip credentials from private PyPi repositories urls.

### Additional Context

Note also https://github.com/conda/conda-lock/issues/460 which proposes improvements to how auth is configured for private PyPi repositories.

PR #470, opened 2023-07-17
Implement auth stripping for private PyPi packages

jacksmith15:
### Description

Related issue: https://github.com/conda/conda-lock/issues/461
Related PRs: https://github.com/conda/conda-lock/pull/323

https://github.com/conda/conda-lock/pull/323 aimed to implement `--strip-auth` support for private PyPi packages, however it only targeted `--kind explicit` lockfiles.

This PR adds support for `--kind lock` lockfiles (the default, unified kind).

To achieve this, I pass the `--strip-auth` flag through to the resolver, and strip HTTP Basic auth from the URLs. This avoids the need for first writing a temporary lockfile containing the credentials, and then fixing them after the fact. It also avoids edge cases where users may have requested multiple types of lockfiles.

Currently only HTTP Basic auth is supported for private PyPi repositories, so this should catch all cases.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -259,6 +259,7 @@ def make_lock_files(  # noqa: C901
     metadata_choices: AbstractSet[MetadataOption] = frozenset(),
     metadata_yamls: Sequence[pathlib.Path] = (),
     with_cuda: Optional[str] = None,
+    strip_auth: bool = False,
 ) -> None:
     """
     Generate a lock file from the src files provided
@@ -385,6 +386,7 @@ def make_lock_files(  # noqa: C901
                 update_spec=update_spec,
                 metadata_choices=metadata_choices,
                 metadata_yamls=metadata_yamls,
+                strip_auth=strip_auth,
             )
 
             if "lock" in kinds:
@@ -685,6 +687,7 @@ def _solve_for_arch(
     platform: str,
     channels: List[Channel],
     update_spec: Optional[UpdateSpecification] = None,
+    strip_auth: bool = False,
 ) -> List[LockedDependency]:
     """
     Solve specification for a single platform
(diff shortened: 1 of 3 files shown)


conda_lock/conda_lock.py
```diff
@@ -1309,16 +1317,17 @@ def lock(
...
             filename_template_temp = f"{tempdir}/{filename_template.split('/')[-1]}"
             lock_func(filename_template=filename_template_temp)
             filename_template_dir = "/".join(filename_template.split("/")[:-1])
             for file in os.listdir(tempdir):
-                lockfile = read_file(os.path.join(tempdir, file))
-                lockfile = _strip_auth_from_lockfile(lockfile)
-                write_file(lockfile, os.path.join(filename_template_dir, file))
+                lockfile_content = read_file(os.path.join(tempdir, file))
+                lockfile_content = _strip_auth_from_lockfile(lockfile_content)
+                write_file(lockfile_content, os.path.join(filename_template_dir, file))
```
jacksmith15, line 1330: A small readability improvement, `lockfile` refers to a file-path earlier in this function's scope.

conda_lock/pypi_solver.py
```diff
@@ -377,3 +381,10 @@ def _prepare_repositories_pool(allow_pypi_requests: bool) -> Pool:
     if allow_pypi_requests:
         repos.append(PyPiRepository())
     return Pool(repositories=[*repos])
+
+
+def _strip_auth(url: str) -> str:
+    """Strip HTTP Basic authentication from a URL."""
+    parts = urlsplit(url, allow_fragments=True)
+    netloc = parts.netloc.replace(f"{parts.username}:{parts.password}@", "")
```
maresb, line 389: I'm worried about this somehow being too clever and missing something, especially since URLs sometimes get encoded. Would this work?

```suggestion
    # Remove everything before and including the last '@' character in the part
    # between 'scheme://' and the subsequent '/'.
    netloc = parts.netloc.split("@")[-1]
```

jacksmith15: That seems sensible 👍

maresb (approved):

PR #471, opened 2023-07-17
Configure private PyPi repositories in the environment

### LINKED ISSUE #460: Improve auth configuration for private PyPi repositories
### Checklist

### What is the idea?

I'd like to improve upon the mechanism for configuring auth for private PyPi repositories.

### Why is this needed?

The documentation claims the process for configuring private pypi repositories is simply:

```shell
poetry config repositories.foo https://username:password@foo.repo/simple/
```

However I believe the process actually looks as follows:

1. Install the correct version `poetry` (matching the vendored version):
    ```shell
    pipx install poetry==1.1.15
    ```
2. Configure repositories using `poetry`:
    ```shell
    poetry config repositories.foo https://username:password@foo.repo/simple/ 
    ```
3. Copy poetry config to location discoverable by `conda-lock`:
    ```shell
    # Note that exact paths vary on different machines
    cp -R ~/.config/pypoetry ~/.config/pypoetry-conda-lock
    ```

I think there are a few problems here:

1. Documentation not up-to-date
2. Users must install `poetry` separately, at a specific version matching the vendored poetry. This negates the value of vendoring in the first place.
3. Users use an interface provided by `poetry`, which is effectively an implementation detail of `conda-lock` (currently `conda-lock` uses `poetry` as a resolver under the hood, but `conda-lock` should be free to change the resolution algorithm without breaking user workflows).
4. Authentication secrets are saved to and copied around on disk.

### What should happen?

I propose that `conda-lock` provides its own interface for configuring private `pypi` repositories, and then manages the internal gymnastics of providing these to the `poetry` resolves.

I propose that this interface supports configuration via environment variable first-and-foremost, as this is the most portable approach for configuration, and supports more secure workflows with tools like [envchain](https://github.com/sorah/envchain).

### Additional Context

Note also separate issue <PLACEHOLDER> regarding auth stripping for private PyPi repositories.

jacksmith15:
### Description

Resolves #460 

This PR adds support for configuring private PyPi repositories via environment variables. This avoids users needing to install an exact version of `poetry` to manage the configuration format, and then copy around credential files to make them discoverable to `conda-lock`. See #460 for full reasoning.

After this change, users are able to set the `CONDA_LOCK_PYPI_REPOSITORY_<repository-name>` environment variable to configure new PyPi repositories.

An alternative could be to support `PIP_EXTRA_INDEX_URL` instead, as this will be respected during `conda-lock install`, meaning only a single environment variable is required for both lock and install.


conda_lock/pypi_solver.py
```diff
@@ -369,6 +371,9 @@ def _prepare_repositories_pool(allow_pypi_requests: bool) -> Pool:
     factory = Factory()
     config = factory.create_config()
     repos = [
+        factory.create_legacy_repository({"name": name, "url": url}, config)
+        for name, url in _parse_repositories_from_environment().items()
+    ] + [
         factory.create_legacy_repository(
             {"name": source[0], "url": source[1]["url"]}, config
         )
```
jacksmith15, line 379: I've left the previous method here so this is not a breaking change.

However since this is exposing an implementation detail as an interface, it might be worth adding a deprecation warning?

maresb: I'm wondering if we're ready yet to deprecate it. I'm very eager to ditch anything Poetry-related, but setting up environment variables can be a pain, and until we get something better in place, some people may prefer the Poetry interface.

conda_lock/pypi_solver.py
```diff
@@ -373,6 +375,9 @@ def _prepare_repositories_pool(allow_pypi_requests: bool) -> Pool:
     factory = Factory()
     config = factory.create_config()
     repos = [
+        factory.create_legacy_repository({"name": name, "url": url}, config)
+        for name, url in _parse_repositories_from_environment().items()
+    ] + [
```
maresb, line 380: Could you please write a test for `_prepare_repositories_pool` to exercise this functionality?

maresb (changes requested):
Looking very good. Let's keep `poetry config repositories.foo` in the docs, but put your method first.

README.md
```diff
@@ -156,13 +156,28 @@ The default category is `main`.
...
+[repositories.example]
+url = "https://username:password@example.repo/simple"
+```
+
+The discoverable location of this file depends on the operating system:
+
+- **Linux**: `$XDG_CONFIG_HOME/pypoetry-conda-lock` or `~/.config/pypoetry-conda-lock`
+- **OSX**:  `~/Library/Application Support/pypoetry-conda-lock`
+- **Windows**: Check output of `python -c 'from conda_lock._vendor.poetry.locations import CONFIG_DIR;print(CONFIG_DIR)'`
+
```
jacksmith15, line 177: @maresb I have restored documentation for the existing repository configuration method here.

Since the previous documentation was incorrect, I have changed the documented process for configuring it this way. This is because the `poetry config repositories.foo` will write configuration to the wrong appdirs directory for `conda-lock` to discover (`~/.config/pypoetry` vs `~/.config/pypoetry-conda-lock`).

I have also written it in a way which avoids users needing to install the exact version of `poetry` vendored by `conda-lock`.

Also note: I couldn't decipher which directory to recommend for windows, as I suspect its platform-dependent, so that's the best I could do. The old documentation didn't have this problem for the same reason it didn't work.

maresb: Thanks a lot for catching this! Seems like we broke this in https://github.com/conda/conda-lock/pull/342.

jacksmith15:
@maresb thanks for the review, and apologies for the delay in updating.

I think I've addressed all of your previous feedback now :+1:

maresb:
Hey @jacksmith15, I feel really bad for doing this, but after taking a fresh look I think we are on the wrong track here. :cry: I should have caught this earlier in the review process, my bad.

I think a much better approach would be a parallel implementation of [authenticated channels](https://conda.github.io/conda-lock/authenticated_channels/) where the corresponding documentation would be essentially identical, but substituting `channels` → `pypi-repositories`.

Do you agree? If so, would you have any time to rework this into a (partial) implementation?

I'm terribly sorry about sending you down the wrong track. Please let me know if I can do anything to help.

jacksmith15:
To be clear, the suggested approach would involve adding a new non-standard key to the source `environment.yml`, e.g. 

```yaml
channels:
  - https://host.tld/t/$QUETZ_API_KEY/channel_name
  - conda-forge
repositories:
  - https://$MY_USER:$MY_PASS@example.repo/pypi/simple
```

Does that look right to you?


Should this also provide a mechanism for enabling/disabling PyPi? Similar to `nodefaults` in channels?

maresb:
These are very good questions. You are understanding my proposal correctly.

We already have a precedent for defining a non-standard key in `environment.yml`, namely `platforms`, so I think adding another should be okay.

I personally would prefer to be more specific with the key name and have it be `pypi-repositories` just because "repository" is such a general term, and I get confused wondering what's Conda-related and what's pip-related. Maybe it should be called `pip-repositories`? (Note `-` characters should be fine for key names in yaml, toml, and command-line arguments.)

I think it's important to avoid Poetry's influence which may push us towards bad design decisions, so I'd prefer autonaming to having multiple conventions. Do you foresee any problems that might arise with this approach?

A mechanism for disabling PyPI outside of `pyproject.toml` and more in line with this system would be interesting, but certainly not necessary at this stage. Probably it would be best to defer this to another PR.

Supporting `--strip-auth` would be nice, but if it's a lot of work I think it'd be fine to instead add a prominent disclaimer.

jacksmith15:
@maresb regarding the naming, its actually a little tricky to give a good name for Python repositories. I think the most "correct" term is "python simple repository": https://packaging.python.org/en/latest/specifications/simple-repository-api/#simple-repository-api

- `pypi` refers specifically to https://pypi.org/
- `pip` refers to the package manager, which is one way to interact with Python Package repositories which conform to the [Simple API](https://packaging.python.org/en/latest/specifications/simple-repository-api/#simple-repository-api)

However `simple-repositories` has a similar issue of being quite vague! I suppose one could do `pep-503-repositories` but that requires additional context.

I'll prefer `pip-repositories` for now as that's probably most familiar to most people, but worth considering which you prefer.

maresb:
@jacksmith15, I agree that I think `pip-repositories` is the best choice.

jacksmith15:
:memo: I haven't forgotten about this!

I am working on the alternative solution in https://github.com/conda/conda-lock/pull/481

There's some differences with pip repositories compared to channels that need handling, but I the general direction seems :+1:

maresb:
Superseded by #529

Issue #474, opened 2023-07-24
InvalidRequirement: Expected package name at the start of dependency specifier


moi90:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The environment is defined as follows:

```yaml
name: base
category: dev
channels:
  - conda-forge
  - defaults
dependencies:
  - make # For docs
  - pip
  - pip:
    - -r docs/requirements.txt
```

Including `requirements.txt` in this way is allowed by conda.
However, with `conda-lock -f environment.base.yml`, I get the following error:

```
Traceback (most recent call last):
  File "/home/mambauser/.local/pipx/shared/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 35, in __init__
    parsed = _parse_requirement(requirement_string)
  File "/home/mambauser/.local/pipx/shared/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 64, in parse_requirement
    return _parse_requirement(Tokenizer(source, rules=DEFAULT_RULES))
  File "/home/mambauser/.local/pipx/shared/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/_parser.py", line 73, in _parse_requirement
    name_token = tokenizer.expect(
  File "/home/mambauser/.local/pipx/shared/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/_tokenizer.py", line 140, in expect
    raise self.raise_syntax_error(f"Expected {expected}")
  File "/home/mambauser/.local/pipx/shared/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/_tokenizer.py", line 165, in raise_syntax_error
    raise ParserSyntaxError(
pkg_resources.extern.packaging._tokenizer.ParserSyntaxError: Expected package name at the start of dependency specifier
    -r docs/requirements.txt
    ^

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/mambauser/.local/bin/conda-lock", line 8, in <module>
    sys.exit(main())
  File "/home/mambauser/.local/pipx/venvs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/mambauser/.local/pipx/venvs/conda-lock/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
[...]

PR #476, opened 2023-07-28
Proposed fix for #474: enables parsing of requirements.txt during locking

stefansjs:
### Description

This addresses https://github.com/conda/conda-lock/issues/474, which fails to lock if environment.yml has -rrequirements.txt in the pip requirements. The main problem with the existing implementation is that it treats requirements lists as if they were a tool.poetry.dependencies block in a pyproject.toml. Poetry is great, and I see why you would want to vendor such a clean implementation, but poetry doesn't have any specification for external files. Further, any differences between calling `pip install` and `poetry lock` would eventually cause an issue.

My fix mimics conda's implementation which writes a temporary requirements.txt and invokes the command-line pip install command. https://github.com/conda/conda/blob/main/conda_env/installers/pip.py. Ideally this means there will be fewer bugs/differences between performing `conda env create` and `conda lock`.

Since pip doesn't provide an explicit library or API for parsing I do it by invoking pip's own parse_requirements function. I also based the return values strongly on the pyproject_toml.py implementation. 

### Testing and Feedback

I would love if somebody could suggest the right place to put unit tests.

### DIFF
--- a/conda_lock/src_parser/environment_yaml.py
+++ b/conda_lock/src_parser/environment_yaml.py
@@ -1,17 +1,15 @@
 import pathlib
 import re
-import sys
-
+import tempfile
 from typing import List, Tuple
 
 import yaml
+from packaging._parser import parse_requirement
 
 from conda_lock.models.lock_spec import Dependency, LockSpecification
 from conda_lock.src_parser.conda_common import conda_spec_to_versioned_dep
+from conda_lock.src_parser.requirements_txt import parse_requirements_txt
 from conda_lock.src_parser.selectors import filter_platform_selectors
-
-from .pyproject_toml import parse_python_requirement
-
 
 _whitespace = re.compile(r"\s+")
 _conda_package_pattern = re.compile(r"^(?P<name>[A-Za-z0-9_-]+)\s?(?P<version>.*)?$")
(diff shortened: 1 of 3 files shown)


conda_lock/src_parser/requirements_txt.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+from packaging._parser import parse_requirement
+from pip._internal.network.session import PipSession
+from pip._internal.req.req_file import parse_requirements
+
+from conda_lock.models.lock_spec import VersionedDependency, VCSDependency, URLDependency, Dependency
+from conda_lock.src_parser.pyproject_toml import unpack_git_url
+
+
+def parse_requirements_txt(file_path, category=None) -> Dependency:
```
maresb, line 12: For the return type annotation, this seems to be `Dependency` generator rather than returning a `Dependency`.

conda_lock/src_parser/environment_yaml.py
```diff
@@ -55,28 +53,18 @@ def _parse_environment_file_for_platform(
...
-                    )
-                )
+            # Generate the temporary requirements file
+            yaml_dir = source_path.parent
+            with tempfile.NamedTemporaryFile(dir=yaml_dir, suffix="requirements.txt", mode='w', encoding='utf-8') as requirements:
+                requirements.writelines(mapping_spec["pip"])
+                requirements.write("\n")  # Trailing newline
+                requirements.file.close()
+
+                dependencies.extend(parse_requirements_txt(requirements.name, category))
```
maresb, line 63: I'm wondering what will happen now with editable requirements.

maresb (commented):
This looks really promising!!! Apologies for taking so long to provide feedback.

I'd recommend putting tests into `tests/test_conda_lock.py`.

I'm a bit uncomfortable tapping into `pip._internal` since that may be an unstable API. (Perhaps it's worth considering if it's feasible to vendor `pip._internal.req.req_file`.)

But overall I think this is a major step in the right direction and a very important feature. Thanks so much for the submission!

PR #481, opened 2023-08-09
Configure private pip repositories in the environment file

jacksmith15:
### Description

This PR adds support for specifying private pip repositories in the `environment.yml` file, similar to how channels are specified.

Similarly to channels, environment variables may be specified, and these will remain as references in the lockfile.

```yaml
channels:
  - conda-forge
pip-repositories:
  - http://$PIP_USER:$PIP_PASSWORD@private-pypi.org/api/pypi/simple
dependencies:
  - python=3.11
  - requests=2.26
  - pip:
    - fake-private-package==1.0.0
```

This aims to solve https://github.com/conda/conda-lock/issues/460, and is an alternative approach to the one proposed by https://github.com/conda/conda-lock/pull/471


conda_lock/models/channel.py
```diff
@@ -1,126 +1,14 @@
-"""
```
jacksmith15, line 1: Most of this has been extracted to `conda_lock/models/package_source.py`, which contains the shared logic common to both channels and pip repositories.

maresb:
Hey, I took a look, and I like the direction this is heading, thanks so much for this!

I'm impressed by your PyPI mock. :)

I see that the first commit is primarily a combination of moving code from `channel.py` to `package_source.py`, pushing through the `pip_repositories` argument, and adding a tiny bit of code to make the `pip_repositories` argument work. But given the large subtractions and additions to `channel.py` and `package_source.py`, it took quite some effort to manually diff everything before I managed to deduce this. It would help me a lot with review if you put any subsequent cut-paste operations into their own commits so that I can distinguish between moved code and altered code.

Finally, feel free to be bold about refactoring and simplifying existing code if it helps. You're touching a lot of code that's probably long overdue for an overhaul. For example, I'd prefer to use [libraries](https://pypi.org/project/packaging/) when it makes things easier.

jacksmith15:
I note an issue with the current approach, which is unfortunately not detected by the test.

The `poetry` solver [makes requests to the repository using `requests`](https://github.com/conda/conda-lock/blob/main/conda_lock/_vendor/poetry/repositories/legacy_repository.py#L385-L410), which appears to automatically swap URL-based Basic Auth with header-based. This is relatively sensible, however it does mean that [this debug warning](https://github.com/conda/conda-lock/blob/main/conda_lock/_vendor/poetry/repositories/legacy_repository.py#L402C7-L408) will almost always fire.

More annoyingly, the solver [uses the URL from the response not the one used to make the request](https://github.com/conda/conda-lock/blob/main/conda_lock/_vendor/poetry/repositories/legacy_repository.py#L410). This means that if the solver makes a request to `https://username:password@private-pypi.org/simple/package`, the resulting URL for the package returned by the solver will be `https://private-pypi.org/simple/package`, which obviously doesn't include the secrets, but also doesn't include the environment variables.

I can only think of one good solution here, which is to inspect each resolved package URL and attempt to match them to the configured `pip-repositories`, and then insert the basic auth back into the URL (as environment variables). The [link objects returned by the solver](https://github.com/conda/conda-lock/blob/main/conda_lock/pypi_solver.py#L217) can be used to match on repositories, because they include [they include a reference to the Page and headers of the request](https://github.com/conda/conda-lock/blob/main/conda_lock/_vendor/poetry/repositories/legacy_repository.py#L122).

I'll investigate this route.

mariusvniekerk:
We should also support parsing credentials dynamically out of netrc files (this works quite well for a lot of conda package registries)

conda_lock/models/pip_repository.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def normalize_solver_url(self, solver_url: str) -> str:
+        """Normalize the URL returned by Poetry's solver.
+
+        Poetry doesn't return URLs with URL-based basic Auth, because it gets converted to
+        header-based Basic Auth. Because of this, we have to add the auth back in here.
+        """
+        if not solver_url.startswith(self.stripped_base_url):
+            # The resolved package URL is at a different host to the repository
+            return solver_url
```
jacksmith15, line 41: This is an interesting edge case.

Basically this is where the repository is at `http://private-pypi.org`, but that index directs the solver to some other host, e.g. `http://packages.private-pypi.org` or `http://somewhere.completely.different`.

Currently in this case the lockfile will just use the URL from the solver as-is, without any auth.

An alternative would be to just add the auth to the URL anyway? WDYT?

jacksmith15:
@maresb @mariusvniekerk 

I (finally) found enough time to get this in a working state, ready-for-review. Let me know what you think.

maresb:
@jacksmith15 I'm doing a lot of traveling at the moment; this is in my review queue, but I'm still looking for a chunk of time when I can go through this carefully. Feel free to pester me, but I can't promise any timescale yet.

jacksmith15:
@maresb I've re-opened this PR in https://github.com/conda/conda-lock/pull/529 with up-to-date changes from `main` and a cleaner git history. Keen to get your thoughts.

Issue #196, opened 2022-05-26
Removing dependencies from an existing lock file


scottyhq:
I'm surprised that *removing* a dependency does not remove it and sub-dependencies from an existing `conda-lock.yml` (condalock 1.0.5):

Steps to reproduce:

1. `conda-lock lock --mamba -f environment.yml` with the following `environment.yml`:
```
channels:
  - conda-forge
dependencies:
  - python=3.9
  - pystac
platforms:
  - linux-64
```

2. Comment or delete the `-pystac` dependency and re-run `conda-lock lock --mamba -f environment.yml`, but `pystac` remains in `conda-lock.yml`:
```
  name: pystac
  url: https://conda.anaconda.org/conda-forge/noarch/pystac-1.4.0-pyhd8ed1ab_0.tar.bz2
```

A workaround is to wipe `conda-lock.yml` and start fresh, but it's nice to keep it around for version control.

riccardoporreca:
Not removing obsolete dependencies is indeed problematic and probably not a desirable behavior.
Even an upgrade of an existing dependency might leave behind former transitive dependencies that might eventually become incompatible with the rest of the current (and updated) dependencies.

weiji14:
Just checking to see if this issue has been resolved in `conda-lock>=2`? Or do we still need to do the `rm conda-lock.yml && conda-lock lock --mamba -f environment.yml` workaround?

maresb:
As far as I know this is unfortunately not yet resolved.

mfisher87:
Today I stumbled on a nasty consequence of this behavior. In my case, I had a dependency in the `pip` dependency subsection of my `environment.yml`, and that dependency became recently available on `conda-forge`. So naturally, I moved it out of the `pip` section and into the main `dependencies` list.

The result was that the same dependency was listed twice, once for `conda` and once for `pip`. Of course, the `pip` version overwrote the `conda` version in the last stage of the install. This resulted in a broken environment with incompatible dependencies, even though our freshly-locked `environment.yml` defined a compatible environment spec.

I created a repository that walks through a simplified version of what I experienced, without the dependency conflicts: https://github.com/mfisher87/sscce-conda-lock-pip-ghosts

[...]

PR #485, opened 2023-08-20
Persist packages from original lockfile _only_ for platforms not requested for lock

itsarobin:
### Description

Addresses #196: for requested platforms replaces lock content without erroneously persisting packages.

TODO:

- [ ] Add unit tests

Co-authored-by: Arie Knoester <arikstr@gmail.com>
Reviewed-by: Matt Fisher <mfisher87@gmail.com>

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -387,7 +387,8 @@ def make_lock_files(
 
         if platforms_to_lock:
             print(f"Locking dependencies for {platforms_to_lock}...", file=sys.stderr)
-            lock_content = lock_content | create_lockfile_from_spec(
+
+            new_lock_content = create_lockfile_from_spec(
                 conda=conda,
                 spec=lock_spec,
                 platforms=platforms_to_lock,
@@ -397,6 +398,21 @@ def make_lock_files(
                 metadata_yamls=metadata_yamls,
                 strip_auth=strip_auth,
             )
+
+            if not lock_content:
+                lock_content = new_lock_content
+            else:
+                # Persist packages from original lockfile for platforms not requested for lock
+                packages_not_to_lock = [
+                    dep
+                    for dep in lock_content.package
+                    if dep.platform not in platforms_to_lock
+                ]
+                filtered_lock_content = lock_content.copy(
+                    deep=True,
+                    update={"package": packages_not_to_lock},
+                )
+                lock_content = filtered_lock_content | new_lock_content
 
             if "lock" in kinds:
                 write_conda_lock_file(
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -250,7 +250,7 @@ def fn_to_dist_name(fn: str) -> str:
     return fn
 
 
-def make_lock_files(
+def make_lock_files(  # noqa: C901
     *,
     conda: PathLike,
     src_files: List[pathlib.Path],
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -341,35 +341,35 @@ def make_lock_files(  # noqa: C901
             virtual_package_repo=virtual_package_repo,
             required_categories=required_categories if filter_categories else None,
         )
-        lock_content: Optional[Lockfile] = None
+        original_lock_content: Optional[Lockfile] = None
 
-        platforms_to_lock: List[str] = []
-        platforms_already_locked: List[str] = []
         if lockfile_path.exists():
             import yaml
 
             try:
-                lock_content = parse_conda_lock_file(lockfile_path)
+                original_lock_content = parse_conda_lock_file(lockfile_path)
             except (yaml.error.YAMLError, FileNotFoundError):
                 logger.warning(
                     "Failed to parse existing lock.  Regenerating from scratch"
                 )
-                lock_content = None
+                original_lock_content = None
         else:
-            lock_content = None
+            original_lock_content = None
 
-        if lock_content is not None:
-            platforms_already_locked = list(lock_content.metadata.platforms)
+        platforms_to_lock: List[str] = []
+        platforms_already_locked: List[str] = []
+        if original_lock_content is not None:
+            platforms_already_locked = list(original_lock_content.metadata.platforms)
             update_spec = UpdateSpecification(
-                locked=lock_content.package, update=update
+                locked=original_lock_content.package, update=update
             )
             for platform in lock_spec.platforms:
                 if (
                     update
-                    or platform not in lock_content.metadata.platforms
+                    or platform not in platforms_already_locked
                     or not check_input_hash
                     or lock_spec.content_hash_for_platform(platform)
-                    != lock_content.metadata.content_hash[platform]
+                    != original_lock_content.metadata.content_hash[platform]
                 ):
                     platforms_to_lock.append(platform)
                     if platform in platforms_already_locked:
@@ -385,10 +385,12 @@ def make_lock_files(  # noqa: C901
             )
         platforms_to_lock = sorted(set(platforms_to_lock))
 
-        if platforms_to_lock:
+        if not platforms_to_lock:
+            new_lock_content = original_lock_content
+        else:
             print(f"Locking dependencies for {platforms_to_lock}...", file=sys.stderr)
 
-            new_lock_content = create_lockfile_from_spec(
+            fresh_lock_content = create_lockfile_from_spec(
                 conda=conda,
                 spec=lock_spec,
                 platforms=platforms_to_lock,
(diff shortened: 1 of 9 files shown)


mfisher87:
@maresb What do you think of this approach? We have to add some unit tests, but want to check that the approach is sound and take feedback in to account before writing unit tests for what might be a bad idea :smile_cat:

conda_lock/conda_lock.py
```diff
@@ -378,7 +378,8 @@ def make_lock_files(  # noqa: C901
 
         if platforms_to_lock:
             print(f"Locking dependencies for {platforms_to_lock}...", file=sys.stderr)
-            lock_content = lock_content | create_lockfile_from_spec(
```
mfisher87, line 381: We believe the original intent of this merge was to persist platforms that were not requested for update (based on [this](https://github.com/conda/conda-lock/commit/6ea3f1cb212ed77216ff27d06630de53847515c9) commit).

We achieve that by only merging the subset of the original `lock_content` that contains those platforms.

conda_lock/conda_lock.py
```diff
@@ -389,6 +390,21 @@ def make_lock_files(  # noqa: C901
...
+            if not lock_content:
+                lock_content = new_lock_content
+            else:
+                # Persist packages from original lockfile for platforms not requested for lock
+                packages_not_to_lock = [
+                    dep
+                    for dep in lock_content.package
+                    if dep.platform not in platforms_to_lock
+                ]
+                filtered_lock_content = lock_content.copy(
```
mfisher87, line 402: Better name: `lock_content_to_persist`?

maresb: Ya. I'd even rename the original to `original_lock_content`. And maybe `new_lock_content` → `fresh_lock_content` in order to reserve `new_lock_content` for the final result.

I'm a big fan of long and descriptive variable names.

conda_lock/conda_lock.py
maresb: Yes, this seems reasonable to me.

conda_lock/conda_lock.py
mfisher87: >

Ya. I'd even rename the original to original_lock_content. And maybe new_lock_content → fresh_lock_content in order to reserve new_lock_content for the final result.
>
>I'm a big fan of long and descriptive variable names.

:100: same here :)

mfisher87:
Unit tests for this change are in. We're thinking one more unit test to validate a single-platform update works as expected (that's not testing new functionality, it's testing a feature that pre-dates this change hasn't regressed).

Then we're thinking we will work on variable naming!

environments/README.md
```diff
@@ -0,0 +1,13 @@
...
+First create the dev environment:
+
+```
+mamba env create -f environments/dev-environment.yaml
+```
+
+Then, add `conda-lock` and its dependencies (as specified in `pyproject.toml`):
+
+```
+pip install -e .
```
mfisher87, line 12: TODO: document "Don't forget to activate the new environment!"

mfisher87:
While doing this renaming work, I felt the need to refactor. I actually got sucked a little bit in to doing a refactor and then pulled myself back and made this change as simple as I could.

This change made it clear that there's a point midway in `make_lock_files` where we switch from operating on old lock content to operating on new lock content (I left a comment to that effect), and that's a good indicator we could extract all the behavior that calculates the new lock content into a new function. There are likely even more functions to extract here, the complexity of this function is pretty high. We could get rid of that `noqa` marker :) My domain knowledge is pretty low, though, so I may need to ask many questions to do such a refactor if you agree it's warranted.

TODO: Rebase again, mark ready for review

conda_lock/conda_lock.py
```diff
@@ -410,7 +428,9 @@ def make_lock_files(
                     file=sys.stderr,
                 )
 
-        assert lock_content is not None
+        # After this point, we're working with `new_lock_content`, never
+        # `original_lock_content` or `fresh_lock_content`.
+        assert new_lock_content is not None
```
mfisher87, line 433: I found this assertion a bit confusing. I was not able to type `new_lock_content: Lockfile` because `original_lock_content` could be `None` when we assign it to `new_lock_content` in the `if not platforms_to_lock` block.

I would like to move this check in to the type system, but that seems better left for a larger refactor effort.

conda_lock/conda_lock.py
```diff
@@ -385,9 +385,12 @@ def make_lock_files(
             )
         platforms_to_lock = sorted(set(platforms_to_lock))
 
-        if platforms_to_lock:
+        if not platforms_to_lock:
+            new_lock_content = original_lock_content
```
mfisher87, line 389: Since previously we were mutating `lock_content` in this conditional block, the "else" condition was lack of mutation. I think making that case explicit and first made this renaming change the most readable.

conda_lock/conda_lock.py
mfisher87: 2a1fb78

mfisher87:
@maresb I feel this is ready for a full review :) Thanks for your time thinking about this change!

maresb:
Thanks so much for this!!!

I'm currently on the road and a bit slammed, so I'm not sure how long it'll take me to give the :+1: but at first glance everything looks good.

I don't like having both `requirements-dev.txt` and `environments/dev-environment.yaml` which are not kept in sync. I think we could delete `requirements-dev.txt` if we simply refactor the `mkdocs` GitHub workflow to use [setup-micromamba](https://github.com/conda/conda-lock/blob/8a71638611a1e39abec736e51ce31b32f159c826/.github/workflows/update-lockfile.yaml#L23-L28). Is this something you'd be able to take care of at the end of this PR? (I'm just thinking that this might be a reasonable add-on since you're already updating the corresponding docs.)

Left: `requirements-dev.txt`, Right: `environments/dev-environment.yaml`
image

mfisher87:
I could probably knock that out! I've used `setup-micromamba` a few times now. I'll see how the day plays out :)

mfisher87:
The test workflow and netlify builds use the `requirements-dev.txt`, so this ended up being a bit broader than expected. Is the netlify build configured in the repo somewhere, or in a GUI? Maybe removing `requirements-dev.txt` should be a different PR?

maresb:
Ah, I'm so sorry, I didn't mean to send you down that rabbit hole!

Let's
1. Undo the deletion of `requirements-dev.txt`
2. Keep mention of `requirements-dev.txt` out of the docs, since I want to delete it
3. Get this merged ASAP

I don't have access to Netlify myself, so it's a black box. We'll need help from @mariusvniekerk on this. I tried setting up Netlify through my GitHub, but this is what I end up with:

image

mfisher87:
>Ah, I'm so sorry, I didn't mean to send you down that rabbit hole!

No sweat at all, I didn't fall in too deep :P I think I've got all the requests done now @maresb!

mfisher87:
@maresb I've used `nwtgck/actions-netlify@v2` before to do Netlify deploys from GHA (both preview deploys and production deploys). With Actions, dealing with PRs from forks becomes complicated due to GitHub's security restrictions. Using `pull_request` event type with a fork PR denies access to repo secrets (needed to push to Netlify), and using `pull_request_target` event type denies access to the changes in the PR, so we can't preview.

One way I've seen to work around this is to have Actions build previews in response to comments (`issue_comment` event type) by people with a particular role, e.g. owner or maintainer, and containing specific text like `/deploy-preview`. Usually those projects will also have a bot that responds to new PRs to tell people how to request a preview, e.g. 

> :robot: Ping @conda-lock-team to request a docs preview. An owner or maintainer must post the comment `/deploy-preview` to trigger.

mfisher87:
```
requests.exceptions.HTTPError: 504 Server Error: Gateway Timeout for url: https://api.anaconda.org/download/conda-forge/micromamba/1.5.1/osx-64/micromamba-1.5.1-0.tar.bz2
```

I'm seeing this error 504 on PR #504 as well (just a coincidence :P )...  I'm able to download this file just fine on my local machine.

maresb:
I'm hoping that the error we're seeing is transient. I'm rerunning the failed tests.

I'm a bit confused by the last commit. Do we really want to offer `requirements-dev.txt` as an option for contributors to set up their dev environments? I see it more as a CI-only dependency.

Apart from that, this is really wonderful, and really boosts the code quality here! As soon as we clarify the README and the failing test I'm ready to merge.

mfisher87:
>I'm hoping that the error we're seeing is transient. I'm rerunning the failed tests.

Looks like we're in luck :partying_face: 

>I'm a bit confused by the last commit

My fault, forgot to push the last one to the correct remote. Forgot I need to be pushing to @itsarobin's fork instead of mine :)

maresb (approved):

maresb:
Thank you @mfisher87! Likewise, it's a pleasure to get such well-curated PRs that I can review in just a few minutes! All this refactoring and cleanup is greatly appreciated.

PR #420, opened 2023-05-27
Add more pre-commit hooks

maresb:
* `no-commit-to-branch` for `main` to enforce feature branches
* pyupgrade to keep style modern
* yesqa to remove redundant noqa comments
* ruff for fast linting
* flake8 plugins
        - flake8-builtins (already installed in dev dependencies)
        - flake8-comprehensions (already installed in dev dependencies)
        - flake8-mutable (already installed in dev dependencies)
        - flake8-use-fstring
        - flake8-use-pathlib

Also corresponding code cleanups to make everything pass.

Finally, eliminate dependencies on `importlib-metadata` and `typing_extensions`.

Issue #483, opened 2023-08-11
Windows CI hangs, fails after 6 hours


maresb:
### What happened?

```
[gw1] [ 51%] PASSED tests/test_conda_lock.py::test_install[mamba-explicit] 
tests/test_conda_lock.py::test_install[mamba-env] 
Error: The operation was canceled.
```

maresb:
We should probably try pinning the mamba version?

maresb:
Thankfully this isn't happening anymore, perhaps due to bugfixes in the latest Mamba?

PR #490, opened 2023-08-28
Add Ruff to pre-commit (before isort + flake8 to test it as a possible replacement)

mfisher87:
### Description

I was doing some development for #485 and felt that flake8 was uncomfortably slow to be running in pre-commit (same with Mypy, but one thing at a time ;) ). I've really loved using Ruff on my other projects so I thought I'd propose swapping over with a minimal PR that aims for consistency with the old ruleset and autofixing a few new rules I felt were uncontroversial based on the changes in #420.


pyproject.toml
```diff
@@ -158,3 +148,45 @@ platforms = ["linux-64", "osx-64", "osx-arm64", "win-64", "osx-arm64", "linux-aa
...
 cachecontrol-with-filecache = ">=0.12.9"
+
+
+[tool.ruff]
+target-version = "py38"
+ignore = [
+    "E501",
+    "F401",
+    "F403",
+    # Disabled during migration to Ruff:
```
mfisher87, line 159: Many of these can be enabled and autofixed, but I felt it would be better to tackle it in a separate PR to keep a reasonable scope for this one :)

pyproject.toml
```diff
@@ -158,3 +148,46 @@ platforms = ["linux-64", "osx-64", "osx-arm64", "win-64", "osx-arm64", "linux-aa
...
+    "C414",
+    "C416",
+    "RUF012",
+    "RUF015",
+]
+line-length = 89
+select = [
+    "A",  # flake8-builtins
+    # "B",  # flake8-bugbear
+    "B006",  # Do not use mutable data structures for argument defaults
```
mfisher87, line 177: `flake8-mutable` errors are caught by this rule. Looks to have detected something that was previously undetected, too.

mfisher87:
PR got autoclosed because of a rebase mistake, woops.

conda_lock/conda_lock.py
```diff
@@ -768,7 +768,7 @@ def create_lockfile_from_spec(
     *,
     conda: PathLike,
     spec: LockSpecification,
-    platforms: List[str] = [],
+    platforms: Optional[List[str]] = None,
```
mfisher87, line 771: Thinking out loud: This is probably safe as-is because this function would only be called one time per CLI invocation. But if that assumption changes, this use of a mutable default value could bite in the future.

requirements-dev.txt
```diff
@@ -1,19 +1,12 @@
 black
```
mfisher87, line 1: If `black` is installed and run with pre-commit, maybe it doesn't need to also be here? Same for mypy?

maresb: It's still useful to have in the dev environment. If you're running VS Code, then black is a requirement for "Format Document..." (although I'm not 100% sure this is still the case since they came out with the Black VS Code extension.)

maresb: Also, ruff is a linter, not a type checker, so I think we want to keep mypy, at least for now.

maresb (commented):
Thanks a lot for this!

I understand mypy being slow, but I've never found flake8 to be cumbersome. Maybe I work more slowly than most people.

I know lots of people are really excited about ruff, and it is incredibly fast. On the other hand, flake8 is tried-and-true, and I like its extensibility with plugins.

How would you feel about leaving in flake8 and isort for now? In my current main project I added ruff at the end of my pre-commit list. That was a mistake because I never see it do anything. Let's do better here and put it at the top of the list, and then if I see flake8 no longer doing anything then I'll learn to trust it.

requirements-dev.txt
mfisher87: > Also, ruff is a linter, not a type checker, so I think we want to keep mypy, at least for now.

Yes, I was more thinking about mypy being on the slow side to use in a pre-commit workflow, not about eliminating it entirely. I don't want to touch mypy for this PR in any case, one thing at a time is enough for me  :laughing:

mfisher87: >It's still useful to have in the dev environment. If you're running VS Code, then black is a requirement for "Format Document..." (although I'm not 100% sure this is still the case since they came out with the Black VS Code extension.)

Interesting. I'm not a VSCode user, but I assumed that would integrate with pre-commit as well. Makes sense!

mfisher87:
I'm OK with whatever you all want to go with :) I noticed flake8 being slow because I'm working sometimes on an older Thinkpad.

To provide my personal experience with Flake8 vs Ruff (and probably sound like a gushing fan, and I am), I've been using flake8 for about 7 years now as a valued daily tool, but have replaced it with Ruff for every project I start or revisit because I truly feel Ruff has far overtaken it in every way. It being fast is just a (really nice) bonus to me.

* I was a big user of flake8 plugins and I was bringing my favorites like bugbear, f-strings, comprehensions, quotes, with my to every project.  When I [migrated QGreenland from flake8 to Ruff](https://github.com/nsidc/qgreenland/pull/534/files), we were using 7 flake8 plugins, and after the migration, we're checking more things with Ruff alone.
* Ruff can autofix almost everything. The amount of time and typing strain this saves me has been life-changing.

The combination of the two above things, more rules + autofixing, means I can do amazing things like change Ruff's `target-version` and have the `U` ruleset automatically migrate my type annotations to the newest way, e.g. `Typing.tuple[...]` -> `tuple[...]`. All I need to do is review it. I've seen it do some surprisingly complicated changes, and it's got my trust at this point (though I still review its changes) :) 

I the best way to try it out is as you described, putting Ruff first but keeping flake8 and isort on board until the team feels they are not adding anything! I'll push a commit to that effect now.

mfisher87:
Unfortunately, isort is reversing a change made by Ruff. I really disagree with `isort`'s choice here :) On the subject, the [isort author endorsed Ruff as a replacement fairly strongly](https://news.ycombinator.com/item?id=34061252).

```patch
diff --git a/conda_lock/lockfile/v2prelim/models.py b/conda_lock/lockfile/v2prelim/models.py
index 2cc35c2..038a7e1 100644
--- a/conda_lock/lockfile/v2prelim/models.py
+++ b/conda_lock/lockfile/v2prelim/models.py
@@ -7,12 +7,10 @@ from conda_lock.lockfile.v1.models import (
     GitMeta,
     HashModel,
     InputMeta,
-    LockMeta,
-    MetadataOption,
-    TimeMeta,
 )
 from conda_lock.lockfile.v1.models import LockedDependency as LockedDependencyV1
 from conda_lock.lockfile.v1.models import Lockfile as LockfileV1
+from conda_lock.lockfile.v1.models import LockMeta, MetadataOption, TimeMeta
 from conda_lock.models import StrictModel
```

Having isort and ruff in the pre-commit chain puts the two tools into a disagreement loop. There may be a way to config them to agree more...

maresb:
Ya, a few months ago there were several days of GitHub incidents where I was observing similar behavior of disappearing pushes.

I'm seeing [95f64ea](https://github.com/conda/conda-lock/pull/490/commits/95f64eafa6724ad5745ac1a7ab0e4f148b8d01ad). @mfisher87 is that your latest? If so, let's get this merged!

maresb (approved):

maresb:
Since when is the Windows test passing? I wonder if that's thanks to the new Mamba/Micromamba releases?

Also, @mfisher87, would you like to prepare a PR with a commit reverting [95f64ea](https://github.com/conda/conda-lock/pull/490/commits/95f64eafa6724ad5745ac1a7ab0e4f148b8d01ad)?

mfisher87:
>Since when is the Windows test passing?

:star_struck: and in less than 30 minutes! That's a big difference from timing out in 6 hours.

>I'm seeing [95f64ea](https://github.com/conda/conda-lock/pull/490/commits/95f64eafa6724ad5745ac1a7ab0e4f148b8d01ad). @mfisher87 is that your latest? If so, let's get this merged!

Yes, looks like GitHub got caught up now :) Sounds great! You're good with using selective ignores to work around the places the tools disagree?

>Also, @mfisher87, would you like to prepare a PR with a commit reverting [95f64ea](https://github.com/conda/conda-lock/pull/490/commits/95f64eafa6724ad5745ac1a7ab0e4f148b8d01ad)?

Sure! Just to make sure I understand, this would be preparing a PR in advance to be merged if/when contributors feel more comfortable with Ruff?

mfisher87:
I felt the `# isort: skip_file` comment needed a bit more context, so I pushed another commit. If isort is removed in the future, it might not be clear that this comment corresponds to the isort library, but has no effect on ruff's isort ruleset (ruff skips look different). I wanted it to be really clear when it can be safely removed.

maresb:
> and in less than 30 minutes! 

Ya, it was only timing out since there was a particular test that was hanging with Mamba, #483. Hopefully I can close it now!

> You're good with using selective ignores to work around the places the tools disagree?

I didn't quite follow this. Are you talking about pre-commit? If so, I didn't set up the ignores for this repo.

> Just to make sure I understand, this would be preparing a PR in advance to be merged if/when contributors feel more comfortable with Ruff?

Exactly.

mfisher87:
>> You're good with using selective ignores to work around the places the tools disagree?
>
>I didn't quite follow this. Are you talking about pre-commit? If so, I didn't set up the ignores for this repo.

I was referring to the new ignores added in this PR - [this](https://github.com/conda/conda-lock/pull/490/files#diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R5-R6) and [this](https://github.com/conda/conda-lock/pull/490/files#diff-96e99a05d6430592d56ce2c74c8b8f28e16199d74640f40ade2ee22e3d8a39efR1). These are a bit hacky, but since we're going to have a reversion PR, we don't need to remember to come back and get rid of them later :)

maresb:
Ya, no worries, that's fine.

PR #493, opened 2023-08-30
Remove flake8 and isort

mfisher87:
In #490, Ruff was added to try it out. Ruff can theoretically replace isort and flake8. This PR removes them and removes compatibility hacks in case we decide we do want to drop isort and flake8.

:heavy_plus_sign: @maresb 

:heart:


maresb:
Thanks!!!

PR #495, opened 2023-08-31
Fix ruff pre-commit

maresb:
@mfisher87, according to the main page for this hook, we need `--exit-non-zero-on-fix`.


mfisher87 (approved):
Great catch, thanks for that :) Totally forgot!

Issue #464, opened 2023-07-15
Attempt to parse the `install` filename as a version 2 lock even when it does not end in `conda-lock.yml`


costrouc:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

When I was using conda-lock I wanted to control the filename that `conda-lock lock ...` produces and used `environment.yml.lock`. I ran into an error when trying to run `conda-lock install .... environment.yml.lock`.

```python
    do_validate_platform(lockfile_contents)
  File "/home/costrouc/.conda/lib/python3.9/site-packages/conda_lock/conda_lock.py", line 180, in do_validate_platform
    platform_lockfile = extract_platform(lockfile)
  File "/home/costrouc/.conda/lib/python3.9/site-packages/conda_lock/conda_lock.py", line 161, in extract_platform
    raise RuntimeError("Cannot find platform in lockfile.")
RuntimeError: Cannot find platform in lockfile.
```

This happens because if the lockfile filename does not currently end in `conda-lock.yml` is parses it as a v1 explicit filename.

### What should happen?

Attempt to parse the lockfile as a version 2 yaml file even when the extension is not `environment.yml.lock`.

### Additional Context

Was at Scipy when I ran into this issue and was talking with @mariusvniekerk about it. Realistically this is a user error but I think this is a confusing issue to run into.

maresb:
Thanks for the suggestion, I agree. Would you like to open a PR?

costrouc:
@maresb yeah would be happy to open one. Will likely be later this week! (Hope that I remember 🤞)

PR #499, opened 2023-09-03
Detect lockfile type based on contents rather than filename

jacksmith15:
### Description

This PR aims to solve https://github.com/conda/conda-lock/issues/464, allowing v2 lockfiles to have arbitrary names. Previously the lockfile renderer would negotiate between `kind=lock` and `kind=explicit` lockfiles based on the `conda-lock.yml` filename suffix.

After this change, the YAML parser is used to distinguish between them. If the lock-file is parse-able YAML, its assumed to be `kind=lock`.

### Notes on documentation

I think some documentation updates are appropriate, most likely under [File naming](https://github.com/conda/conda-lock/blob/main/README.md#file-naming), however I'm a bit confused by how this is currently documented:

> The extension `.conda-lock.yml` will be added if not present.

I don't think this is currently accurate, and can probably just be removed as part of this change?

> Rendered environment files (env or explicit) must end with `.lock` and will be named as
`"conda-{platform}.lock"` by default.

I don't this is correct for `env`, and I'm not sure its actually validated for `explicit` either.

Resolves: https://github.com/conda/conda-lock/issues/464

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,7 @@ and update operations. You can supply a different filename with e.g.
 conda-lock --lockfile superspecial.conda-lock.yml
 ```
 
-The extension `.conda-lock.yml` will be added if not present. Rendered
-environment files (env or explicit) must end with `.lock` and will be named as
-`"conda-{platform}.lock"` by default.
+Rendered `explicit` and `env` lockfiles will be named as `"conda-{platform}.lock"` and `"conda-{platform}.lock.yml` respectively by default.
 
 If you want to override that call conda-lock as follows.
 ```bash
@@ -89,7 +87,7 @@ conda-lock -k explicit --filename-template "specific-{platform}.conda.lock"
 Conda-lock will build a spec list from several files if requested.
 
 ```bash
-conda-lock -f base.yml -f specific.yml -p linux-64 --filename-template "specific-{platform}.lock"
+conda-lock -f base.yml -f specific.yml -p linux-64 -k explicit --filename-template "specific-{platform}.lock"
 ````
 
 In this case all dependencies are combined, and the ordered union of all `channels` is used as the final
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -54,7 +54,11 @@ from conda_lock.invoke_conda import (
     determine_conda_executable,
     is_micromamba,
 )
-from conda_lock.lockfile import parse_conda_lock_file, write_conda_lock_file
+from conda_lock.lockfile import (
+    UnknownLockfileVersion,
+    parse_conda_lock_file,
+    write_conda_lock_file,
+)
 from conda_lock.lockfile.v2prelim.models import (
     GitMeta,
     InputMeta,
@@ -933,13 +937,14 @@ def _render_lockfile_for_install(
 
     """
 
-    if not filename.name.endswith(DEFAULT_LOCKFILE_NAME):
+    try:
+        lock_content = parse_conda_lock_file(pathlib.Path(filename))
+    except (yaml.YAMLError, UnknownLockfileVersion):
+        # This indicates a kind explicit lockfile, which is already rendered
         yield filename
         return
 
     from ensureconda.resolve import platform_subdir
-
-    lock_content = parse_conda_lock_file(pathlib.Path(filename))
 
     platform = platform_subdir()
     if platform not in lock_content.metadata.platforms:
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -36,6 +36,7 @@ import click
 import yaml
 
 from ensureconda.api import ensureconda
+from ensureconda.resolve import platform_subdir
 from typing_extensions import Literal
 
 from conda_lock.click_helpers import OrderedGroup
@@ -946,8 +947,6 @@ def _render_lockfile_for_install(
         return
 
     lock_content = parse_conda_lock_file(pathlib.Path(filename))
-
-    from ensureconda.resolve import platform_subdir
 
     platform = platform_subdir()
     if platform not in lock_content.metadata.platforms:
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -143,6 +143,10 @@ _implicit_cuda_message = """
 """
 
 
+class UnknownLockfileKind(ValueError):
+    pass
+
+
 def _extract_platform(line: str) -> Optional[str]:
     search = PLATFORM_PATTERN.search(line)
     if search:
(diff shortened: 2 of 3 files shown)


maresb:
Thanks so much for this! This will be a big improvement.

I agree that the docs are wrong. Would you be able to remove the incorrect info and make any additions, if any are needed?

Regarding the exception handling, as I understand, there's a possibility that an explicit file parses as valid YAML. (Do you have a concrete example of this?) In this case we would get `version=None`, leading to a `UnknownLockfileVersion` exception. When detecting the file type, it seems to me like we should assume `.lock` when `version=None`, but I think we should not assume lock if `version` exists and is not recognized. Thus I think we should add a `VersionMissingFromLockfile` exception so we can catch `VersionMissingFromLockfile` or `yaml.YAMLError`. Does this make sense?

jacksmith15:
> Regarding the exception handling, as I understand, there's a possibility that an explicit file parses as valid YAML.

@maresb I actually added the `UnknownLockVersion` to have `kind=env` lockfiles, which of course are valid yaml. It was caught by the tests.

I can certainly switch to a "version missing" exception.

maresb:
@jacksmith15, yes, and I really like what you did with `UnknownLockfileVersion`. I'm not suggesting switching this to `VersionMissing`, but rather adding `VersionMissing` in order to accurately handle distinct cases: `version=None` with `VersionMissing` and `version=100` with `UnknownLockfileVersion`.

maresb:
I was a bit confused before, so for reference let me lay it out here. We need to deal with the three different types of lockfile listed below (`explicit`, `env`, unified `lock`file).

From the main description above:
> If the lock-file is parse-able YAML, its assumed to be `kind=lock`. 

My confusion before was that I forgot about `env` lockfiles, since these are also parseable YAML. Therefore, this assumption that parseable ⇒ `lock` doesn't seem accurate.

It seems to me like the logic should be something like:

* If the file contains the line `@EXPLICIT` then it's type `explicit`
* Otherwise the file should be parseable YAML, if not raise an exception.
  * If the file contains top-level keys `channels` and `dependencies`, the file is type `env`.
  * If the file contains the top-level key-value pair `version=1` or (TODO for Lockfile v2?) a top-level key named `version` and `metadata.description='unified lockfile'`, then the file is type `lock`.
  * Otherwise raise an exception

`explicit`:
```
# Generated by conda-lock.
# platform: linux-64
# input_hash: f75fe5e8f997bb28184f80929c3dc76a1b25577cd6bb6ee2d789c47b9cd49c55
@EXPLICIT
https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2#d7c89558ba9fa0495403155b64376d81
https://conda.anaconda.org/conda-forge/linux-64/libgomp-13.1.0-he5830b7_0.conda#56ca14d57ac29a75d23a39eb3ee0ddeb
https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2#73aaf86a425cc6e73fcf236a5a46396d
https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.1.0-he5830b7_0.conda#cd93f779ff018dd85c7544c015c9db3c
https://conda.anaconda.org/conda-forge/linux-64/libzlib-1.2.13-hd590300_5.conda#f36c115f1ee199da648e0597ec2047ad
https://conda.anaconda.org/conda-forge/linux-64/zlib-1.2.13-hd590300_5.conda#68c34ec6149623be41a1933ab996a209
```


`env`:

```yaml
# Generated by conda-lock.
# platform: linux-64
# input_hash: 2a6830d8ed0f67005556f1a22c7e1a97d350a9fb5c91e0dbba4b01e0685eccc5

channels:
  - conda-forge
  - defaults
dependencies:
  - _libgcc_mutex=0.1=conda_forge
  - libgomp=13.1.0=he5830b7_0
  - _openmp_mutex=4.5=2_gnu
  - libgcc-ng=13.1.0=he5830b7_0
  - libzlib=1.2.13=hd590300_5
  - zlib=1.2.13=hd590300_5
```


unified `lock`file:
```yaml
#
#
#     conda-lock lock  --lockfile conda-lock.yml --update PACKAGE
#     conda-lock -f environment.yaml --lockfile conda-lock.yml
version: 1
metadata:
  content_hash:
    linux-64: f75fe5e8f997bb28184f80929c3dc76a1b25577cd6bb6ee2d789c47b9cd49c55
  channels:
  - url: conda-forge
    used_env_vars: []
  platforms:
  - linux-64
  sources:
  - environment.yaml
package:
- name: _libgcc_mutex
  version: '0.1'
  manager: conda
  platform: linux-64
  dependencies: {}
  url: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
  hash:
    md5: d7c89558ba9fa0495403155b64376d81
    sha256: fe51de6107f9edc7aa4f786a70f4a883943bc9d39b3bb7307c04c41410990726
  category: main
  optional: false
- name: libgomp
  version: 13.1.0
  manager: conda
  platform: linux-64
  dependencies:
    _libgcc_mutex: '0.1'
  url: https://conda.anaconda.org/conda-forge/linux-64/libgomp-13.1.0-he5830b7_0.conda
  hash:
    md5: 56ca14d57ac29a75d23a39eb3ee0ddeb
    sha256: 5d441d80b57f857ad305a65169a6b915d4fd6735cdc9e9bded35d493c91ef16d
  category: main
  optional: false
- name: _openmp_mutex
  version: '4.5'
  manager: conda
  platform: linux-64
  dependencies:
    _libgcc_mutex: '0.1'
    libgomp: '>=7.5.0'
  url: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
  hash:
    md5: 73aaf86a425cc6e73fcf236a5a46396d
    sha256: fbe2c5e56a653bebb982eda4876a9178aedfc2b545f25d0ce9c4c0b508253d22
  category: main
  optional: false
- name: libgcc-ng
  version: 13.1.0
  manager: conda
  platform: linux-64
  dependencies:
    _libgcc_mutex: '0.1'
    _openmp_mutex: '>=4.5'
  url: https://conda.anaconda.org/conda-forge/linux-64/libgcc-ng-13.1.0-he5830b7_0.conda
  hash:
    md5: cd93f779ff018dd85c7544c015c9db3c
    sha256: fba897a02f35b2b5e6edc43a746d1fa6970a77b422f258246316110af8966911
  category: main
  optional: false
- name: libzlib
  version: 1.2.13
  manager: conda
  platform: linux-64
  dependencies:
    libgcc-ng: '>=12'
  url: https://conda.anaconda.org/conda-forge/linux-64/libzlib-1.2.13-hd590300_5.conda
  hash:
    md5: f36c115f1ee199da648e0597ec2047ad
    sha256: 370c7c5893b737596fd6ca0d9190c9715d89d888b8c88537ae1ef168c25e82e4
  category: main
  optional: false
- name: zlib
  version: 1.2.13
  manager: conda
  platform: linux-64
  dependencies:
    libgcc-ng: '>=12'
    libzlib: 1.2.13
  url: https://conda.anaconda.org/conda-forge/linux-64/zlib-1.2.13-hd590300_5.conda
  hash:
    md5: 68c34ec6149623be41a1933ab996a209
    sha256: 9887a04d7e7cb14bd2b52fa01858f05a6d7f002c890f618d9fcd864adbfecb1b
  category: main
  optional: false
```

jacksmith15:
Thanks for laying that out @maresb ! I considered trying to detect based on content but wasn't sure exactly which rules could be relied on. That makes it very straightforward, thanks.

maresb:
Great! I hope I'm right; I guess we'll see. I really appreciate your openness on this.

jacksmith15:
@maresb I've updated the PR based on your input :pray:

maresb (approved):
Thanks a lot! This looks great to me.

Issue #515, opened 2023-09-27
Install development dependencies incrementally


nealstewart:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

I use conda-lock to lock versions before building Docker images, and also to create a local development environment. With conda-lock's development dependencies, I can create a _fresh_ conda environment with said dependencies, but I would like to be able to update an existing conda environment seamlessly to do so _and include pip dependencies that are locked_. 

### Why is this needed?

Essentially this would just support staged Docker builds without adding egregious amounts of time to my build in order to install with conda-lock twice.

### What should happen?

```
conda-lock install --no-dev -n example
conda-lock install --dev -n example # Incrementally installs the development. dependencies
```

maresb:
I use it similarly and would really like to see this. In particular it would be great if `conda-lock install` didn't clobber existing environments, but rather diffed with the existing dependencies in order to figure out the necessary changes.

Another potentially nice usage case along these lines with Docker would be, in the case of large environments, to apply sequences of lockfiles in order to avoid a full rebuild, e.g.

```Dockerfile
COPY conda-lock-01.yml .
conda-lock install conda-lock-01.yml
COPY conda-lock-02.yml .
conda-lock install conda-lock-02.yml
```

@nealstewart, would you be willing to contribute a PR?

nealstewart:
@maresb PR incoming...

nealstewart:
I've implemented it as an optional new flag `--update`, in order to preserve the existing API behaviour (install twice, create two environments).

I just leverage `conda env update` behind the scenes to leverage Conda's existing behaviour.

nealstewart:
There you go!

sjdemartini:
As mentioned in https://github.com/conda/conda-lock/pull/516#issuecomment-2059418711, this would be useful not just for development dependencies, but it would be super useful in general to be able to update an existing env to match a lockfile, rather than deleting/recreating the env entirely as seems to happen now. The latter approach is unsafe for any shared environments (e.g. a JupyterHub kernel env), but conda-lock would be very useful there for updating environments predictably.

PR #516, opened 2023-09-28
Proposed implementation for #515: Install development dependencies incrementally

nealstewart:
…

### Description

See [issue 515](https://github.com/conda/conda-lock/issues/515 for background on why this was implemented.

I have implemented this as an optional `--update` flag to be based to `conda-lock install`.

This allows for the following use-case, which will support staged Docker builds, e.g. for light-weight testing images:

```
conda-lock install
conda-lock install --dev --update
```

I've copied the existing test and there's some potential for deduplication, but I've gone for simplicity. More than happy to refactor the two highly-coupled tests a bit.

     Helpful links:
       - Conda Org COC: https://github.com/conda-incubator/governance/blob/main/CODE_OF_CONDUCT.md
       - Contributing docs: ../blob/main/CONTRIBUTING.md -->


maresb:
This is exciting! Looks fairly straightforward too. Could you please look into and fix these test failures?

I'm a bit overloaded at the moment, so no promises on timelines for getting this merged, but I'm hoping it will be easy once the CI is green.

sjdemartini:
I'd love to see this change land—it would be super useful to be able to update an existing env to match a lockfile, rather than deleting/recreating the env as seems to happen now. The latter approach is unsafe for any shared environments (e.g. a JupyterHub kernel env), but conda-lock would be very useful there for updating environments predictably.

PR #526, opened 2023-10-12
Parses `python` version for `poetry`'s `pyproject.toml`, always managed by `conda`

pep-sanwer:
This resolves the oversight uncovered at the tail end of [Default PyPI dependencies for pyproject.toml #334](https://github.com/conda/conda-lock/pull/362).
`Poetry`'s `pyproject.toml` python version is now parsed and always managed by `conda` (as it is not a dependency that can be managed by `pip`)

Added [ Poetry's required Python version](https://python-poetry.org/docs/pyproject/) (search `declaring the python version`) specification for all `poetry` `pyproject.toml` tests

TODO: implement for non-poetry `pyproject.toml` flavors (pdm, filt) - a bit complicated as they have differing implementations / locations for python version specification.

### DIFF
--- a/conda_lock/src_parser/pyproject_toml.py
+++ b/conda_lock/src_parser/pyproject_toml.py
@@ -227,6 +227,11 @@ def parse_poetry_pyproject_toml(
             extras: List[Any] = []
             in_extra: bool = False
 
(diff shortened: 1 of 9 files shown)


maresb (approved):
Thanks @pep-sanwer!!!

PR #465, opened 2023-07-15
add osx-arm64 as a default platform

mariusvniekerk:

PR #527, opened 2023-10-12
New Feature: lock only `conda-lock` dependencies in `pyproject.toml`

pep-sanwer:
This introduces a new flag for `pyproject.toml`-based specification (affecting all flavors of `pyproject.toml`) that skips locking / solving for all dependencies not found under `[tool.conda-lock]`.  

This allows use-cases where a user wants `conda-lock` locking for only conda dependencies, while bringing their own locker for other dependencies. (My preferred workflow is `conda-lock` for venv management / management of any non-pip-able deps, followed by `poetry` for pip-able deps).

The flag is exposed as:
```toml
[tool.conda-lock]
skip-non-conda-lock = true
```

The below example `pyproject.toml`:
```toml
[tool.poetry]
name = "conda-lock-test-poetry"
version = "0.0.1"
description = ""
authors = ["conda-lock"]

[tool.poetry.dependencies]
python = "^3.7"
pandas = "*"
numpy = "*"

[build-system]
requires = ["poetry>=0.12"]
build-backend = "poetry.masonry.api"

[tool.conda-lock]
channels = [
    'defaults'
]
skip-non-conda-lock = true

[tool.conda-lock.dependencies]
python = "3.11"
sqlite = "<3.34"
certifi = ">=2019.11.28"
```
would result in `conda-lock` only producing a lockfile for `python`, `sqlite`, and `certifi`, skipping `pandas` and `numpy`.

Again, deeply appreciative of this incredible tool, and for your time in reviewing this PR. Very open to suggestions / refinements.

### DIFF
--- a/tests/test-flit-default-pip/pyproject.toml
+++ b/tests/test-flit-default-pip/pyproject.toml
@@ -8,6 +8,7 @@ requires = [
     "toml >=0.10",
     "tomlkit >=0.7",
     ]
+requires-python = ">=3.7"
 
 [tool.flit.metadata.requires-extra]
 test = [
@@ -25,5 +26,5 @@ channels = [
 default-non-conda-source = "pip"
 
 [tool.conda-lock.dependencies]
-sqlite = "<3.34"
+sqlite = ">=3.34"
 certifi = ">=2019.11.28"
--- a/README.md
+++ b/README.md
@@ -359,6 +359,7 @@ the following sections to the `pyproject.toml`
 [tool.conda-lock.dependencies]
 sqlite = ">=3.34"
 ```
+
 #### pip dependencies
 
 If a dependency refers directly to a URL rather than a package name and version,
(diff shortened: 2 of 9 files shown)


maresb (changes requested):
This is brilliant! Your use case is compelling, and the change is nice and simple.

This looks like it should automatically work for non-poetry build backends. Would you be able to copy and tweak this test with your new `skip-non-conda-lock` flag?
https://github.com/conda/conda-lock/blob/425b384ffd010461d9a4f3c61d286e31a21f14f3/tests/test_conda_lock.py#L779-L796

Once this is done, it looks good to me.

tests/test-flit-default-pip/pyproject.toml
```diff
@@ -25,5 +26,5 @@ channels = [
 default-non-conda-source = "pip"
 
 [tool.conda-lock.dependencies]
-sqlite = "<3.34"
+sqlite = ">=3.34"
```
maresb, line 29: What's the reason for this? It seems strange to me.

maresb: I just want to make sure I understand what's going on.

I looked into it, and what's happening is that your command is pulling from Anaconda's `defaults` channel, and the first build there of `sqlite` for `osx-arm64` is v3.35.4. We recently added `osx-arm64` as a default platform in #465. Thus the requested environment is indeed not solvable.

I'm happy to update the `sqlite` pins so that they're solvable with the `defaults` channel for `osx-arm64`.

maresb (approved):

PR #531, opened 2023-10-16
Simplify suffix_union function

### LINKED ISSUE #530: Why is suffix_union so complicated?
### What happened?

I'm having a very difficult time understanding how `suffix_union` works. It seems to me like it should be simpler.

Maybe first we should define `prefix_union`, and then `suffix_union` would be the reverse of `prefix_union` with all the arguments reversed.

Now I think `prefix_union` should be:
* Loop over n=0...inf
* Take the n-th element from each sequence that has an n-th element
* In case no sequence contains an n-th element, terminate with success
* Check that all the n-th elements are all equal
* Append the n-th element to the result and continue looping

In particular, I don't see why we should have `SupportsRichComparisonT`.

Context: I'm trying to review #529.

Attn: @mariusvniekerk, @jacksmith15

maresb:
### Description

Closes #530. Let's see if the tests pass...

### DIFF
--- a/conda_lock/common.py
+++ b/conda_lock/common.py
@@ -79,9 +79,7 @@ def ordered_union(collections: Iterable[Iterable[T]]) -> List[T]:
     return list({k: k for k in chain.from_iterable(collections)}.values())
 
 
-def suffix_union(
-    collections: Iterable[Sequence["SupportsRichComparisonT"]],
-) -> List["SupportsRichComparisonT"]:
+def suffix_union(collections: Iterable[Sequence]) -> List:
     """Generates the union of sequence ensuring that they have a common suffix.
 
     This is used to unify channels.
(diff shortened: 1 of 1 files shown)


maresb:
Attn @jacksmith15

conda_lock/common.py
```diff
@@ -92,25 +90,31 @@ def suffix_union(
...
-    RuntimeError: [4, 1] is not a subset of [2, 1]
+    ValueError: [4, 1] is not an ordered subset of [2, 1]
 
     """
-    from genericpath import commonprefix
+    return list(reversed(prefix_union([list(reversed(s)) for s in collections])))
 
-    result: List["SupportsRichComparisonT"] = []
+
+def prefix_union(collections: Iterable[Sequence]) -> List:
```
jacksmith15, line 99: I wonder if part of the confusion has stemmed from the generic naming of these functions.

Could it be instead called `unify_package_sources` or `merge_package_sources`?

FWIW the way I reason about this logic is a bit different, not sure if this is more or less confusing to others (probably a tad less performant):

```python
def unify_package_sources(collections: Iterable[Sequence[str]]) -> List[str]:
    result = max(collections, key=len)
    for collection in collections:
        if collection != result[-len(collection):]:
            raise ValueError(f"{collection} is not an ordered subset of {result}")
    return result
```

maresb: Nice!! I think there's an edge case when `len(collection) == 0`. But making sure that's in order, would you like to make a PR to supersede mine?

jacksmith15: Happy to do so :+1:

maresb:
Superseded by #532

PR #532, opened 2023-10-16
Replace `suffix_union` with less general function

### LINKED ISSUE #530: Why is suffix_union so complicated?

jacksmith15:
### Description

An alternative candidate to #531 

This removes the quite complex `suffix_union` function with a more specific function for unifying channels when aggregating lock specifications, since that's the only place we use it.

If nothing else, this lets us document why we're doing it better since the function pertains specifically to merging packages sources.

Resolves #530

### DIFF
--- a/conda_lock/src_parser/aggregation.py
+++ b/conda_lock/src_parser/aggregation.py
@@ -3,8 +3,9 @@ import logging
 from itertools import chain
 from typing import Dict, List, Tuple
 
-from conda_lock.common import ordered_union, suffix_union
+from conda_lock.common import ordered_union
 from conda_lock.errors import ChannelAggregationError
+from conda_lock.models.channel import Channel
 from conda_lock.models.lock_spec import Dependency, LockSpecification
 
 
@@ -37,7 +38,9 @@ def aggregate_lock_specs(
         dependencies[platform] = list(unique_deps.values())
 
     try:
-        channels = suffix_union(lock_spec.channels for lock_spec in lock_specs)
+        channels = unify_package_sources(
+            [lock_spec.channels for lock_spec in lock_specs]
+        )
     except ValueError as e:
         raise ChannelAggregationError(*e.args)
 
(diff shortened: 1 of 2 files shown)


conda_lock/src_parser/aggregation.py
```diff
@@ -51,3 +54,34 @@ def aggregate_lock_specs(
             lock_spec.allow_pypi_requests for lock_spec in lock_specs
         ),
     )
+
+
+def unify_package_sources(collections: List[List[Channel]]) -> List[Channel]:
```
jacksmith15, line 59: In #529 I would replace `Channel` here with:

```python
PackageSource = TypeVar("PackageSource", Channel, PipRepository)
```

conda_lock/src_parser/aggregation.py
```diff
@@ -51,3 +54,34 @@ def aggregate_lock_specs(
...
+    > unify_package_sources([[channel_two, channel_one], [channel_three, channel_one]])
+
+    In the failing example, it is not possible to predictably decide which channel
+    to search first, `channel_two` or `channel_three`, so we error in this case.
+    """
+    if not collections:
+        return []
+    result = max(collections, key=len)
+    for collection in collections:
+        if collection != result[-len(collection) :]:
```
maresb, line 85: This still suffers from the edge case I mentioned earlier:

```
unify_package_sources([[], [1,2,3]])

ValueError: [] is not an ordered subset of [1, 2, 3]
```

because `result[-0 :]` is the same as `result[0:]` which is the whole list.

maresb:
We should also be able to remove `__lt__` from `Channel`

conda_lock/src_parser/aggregation.py
```diff
@@ -51,3 +54,34 @@ def aggregate_lock_specs(
...
+
+    In the failing example, it is not possible to predictably decide which channel
+    to search first, `channel_two` or `channel_three`, so we error in this case.
+    """
+    if not collections:
+        return []
+    result = max(collections, key=len)
+    for collection in collections:
+        if collection and collection != result[-len(collection) :]:
+            raise ValueError(f"{collection} is not an ordered subset of {result}")
```
maresb, line 86: I like code that I don't have to think about. While not as slick as your solution, I think it's a bit more explicit.

Also, I think the `ValueError` message was wrong (missing "at the end of").

```suggestion
        if collection == []:
            truncated_result = []
        else:
            truncated_result = result[-len(collection) :]
        if collection != truncated_result:
            raise ValueError(
                f"{collection} is not an ordered subset at the end of {result}"
            )
```

maresb (approved):
@mariusvniekerk I hope this is okay, I'm going to merge it

maresb:
Thanks so much @jacksmith15, I really appreciate all the effort here! I think this is a huge improvement.

PR #535, opened 2023-10-21
Fix platform validation for non-default names

jacksmith15:
### Description

https://github.com/conda/conda-lock/pull/499 added logic to support lockfiles with arbitrary names, by detecting the lockfile kind based on contents rather than name.

However it missed some logic for deciding whether to validate the platform, which still infers lock-file kind from the filename.

This leads to the following error when installing from a lock-file with non-default name:

```
RuntimeError: Cannot find platform in lockfile.
```

This can be worked around by passing `--no-validate-platform`, but is not the intended behaviour.


maresb:
Thanks! Just cut a release for this.

PR #549, opened 2023-11-10
feat: add pypi-to-conda-name overrides to pyproject parsing

### LINKED ISSUE #548: Feature: provide local additions to grayskull_pypi_mapping.yaml
### Checklist

### What is the idea?

Apologies if I've missed this feature in the readme or prior discussion in the issues.

I'm helping someone setup conda-lock with a pyproject.toml setup that includes:

```toml
[project.optional-dependencies]
cuda = [
    "cupy-cuda11x ; platform_system != 'Darwin'",
]
```

naturally, specific cuda builds of of cupy aren't declared on conda [as they are on pypi](https://pypi.org/project/cupy-cuda11x/), but grayskull's lookup doesn't (currently) deal with all of those maps.

I know that I can provide my own `--pypi_to_conda_lookup_file`, but that appears to A) require a URI I can easily pass to requests.get B) presumably requires me to duplicate stuff that I need from grayskull's map?

looking through the source code, there doesn't appear to be any way to append to the lookup.  

### Why is this needed?

to assist in pypi-to-conda name mapping with minimal duplication

### What should happen?

I'm thinking something like:

```toml
[tool.conda-lock.pypi-to-conda-name]
cupy-cuda11x = {"conda_name" = "cupy"}
```

or simpler if no additional info is needed
```toml
[tool.conda-lock.pypi-to-conda-name]
cupy-cuda11x = "cupy"
```

### Additional Context

_No response_

tlambert03:
### Description

Closes #548  by adding support for a `pypi-to-conda-name`  section in pyproject.toml

```toml
[tool.conda-lock.pypi-to-conda-name]
cupy-cuda11x = "cupy"
```

names present there will override mappings provided by grayskull

Note: in #548, I proposed accepting a string or list of strings (which I do think is a good idea), but the implications of accepting a list of strings breaks the current 1-to-1 mapping behavior between pypi and conda, so I think that's probably best handled in a followup PR.  This PR only accepts single string

### DIFF
--- a/conda_lock/lookup.py
+++ b/conda_lock/lookup.py
@@ -1,19 +1,19 @@
 from collections import ChainMap
 from contextlib import suppress
 from functools import cached_property
-from typing import Dict, Mapping, Optional, Union
+from typing import Dict, Mapping, Optional, Union, cast
 
 import requests
 import yaml
 
 from packaging.utils import NormalizedName, canonicalize_name
-from typing_extensions import TypedDict
+from typing_extensions import NotRequired, TypedDict
 
 
 class MappingEntry(TypedDict):
     conda_name: str
     # legacy field, generally not used by anything anymore
-    conda_forge: str
+    conda_forge: NotRequired[str]
     pypi_name: NormalizedName
 
 
@@ -64,8 +64,8 @@ class _LookupLoader:
                     raise ValueError(
                         "MappingEntries must have both a 'conda_name' and 'pypi_name'"
                     )
-                entry = dict(v)
-                entry["pypi_name"] = canonicalize_name(entry["pypi_name"])
+                entry = cast("MappingEntry", dict(v))
+                entry["pypi_name"] = canonicalize_name(str(entry["pypi_name"]))
             elif isinstance(v, str):
                 entry = {"conda_name": v, "pypi_name": key}
             else:
@@ -78,7 +78,7 @@ class _LookupLoader:
     @property
     def pypi_lookup(self) -> Mapping[NormalizedName, MappingEntry]:
         """ChainMap of PyPI to conda name mappings.
-        
+
         Local mappings take precedence over remote mappings fetched from `_mapping_url`.
         """
         return ChainMap(self.local_mappings, self.remote_mappings)
@@ -91,7 +91,7 @@ class _LookupLoader:
 LOOKUP_OBJECT = _LookupLoader()
 
 
-def get_forward_lookup() -> Dict[NormalizedName, MappingEntry]:
+def get_forward_lookup() -> Mapping[NormalizedName, MappingEntry]:
     global LOOKUP_OBJECT
     return LOOKUP_OBJECT.pypi_lookup
 
@@ -112,7 +112,8 @@ def set_lookup_location(lookup_url: str) -> None:
 def set_pypi_lookup_overrides(mappings: Mapping[str, Union[str, MappingEntry]]) -> None:
     """Set overrides to the pypi lookup"""
     global LOOKUP_OBJECT
-    LOOKUP_OBJECT.local_mappings = mappings
+    # type ignore because the setter will normalize the types
+    LOOKUP_OBJECT.local_mappings = mappings  # type: ignore [assignment]
 
 
 def conda_name_to_pypi_name(name: str) -> NormalizedName:
--- a/conda_lock/lookup.py
+++ b/conda_lock/lookup.py
@@ -1,5 +1,7 @@
+from collections import ChainMap
+from contextlib import suppress
 from functools import cached_property
-from typing import Dict
+from typing import Dict, Mapping, Optional, Union
 
 import requests
 import yaml
@@ -17,6 +19,7 @@ class MappingEntry(TypedDict):
 
 class _LookupLoader:
     _mapping_url: str = "https://raw.githubusercontent.com/regro/cf-graph-countyfair/master/mappings/pypi/grayskull_pypi_mapping.yaml"
+    _local_mappings: Optional[Dict[NormalizedName, MappingEntry]] = None
 
     @property
     def mapping_url(self) -> str:
@@ -24,12 +27,16 @@ class _LookupLoader:
 
     @mapping_url.setter
     def mapping_url(self, value: str) -> None:
-        del self.pypi_lookup
-        del self.conda_lookup
+        # these will raise AttributeError if they haven't been cached yet.
+        with suppress(AttributeError):
+            del self.remote_mappings
+        with suppress(AttributeError):
+            del self.conda_lookup
         self._mapping_url = value
 
     @cached_property
-    def pypi_lookup(self) -> Dict[NormalizedName, MappingEntry]:
+    def remote_mappings(self) -> Dict[NormalizedName, MappingEntry]:
+        """PyPI to conda name mapping fetched from `_mapping_url`"""
         res = requests.get(self._mapping_url)
         res.raise_for_status()
         lookup = yaml.safe_load(res.content)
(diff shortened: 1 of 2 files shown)


tlambert03:
ready for a first pass @maresb

tlambert03:
i didn't yet add a command line option, but could also do that if you want to tell me the syntax you'd prefer.  perhaps something like `--pypi-to-conda=cupy-cuda11x:cupy`?  dunno... could get hairy :)

maresb:
> i didn't yet add a command line option

A CLI option doesn't seem to me like a good interface. If you have pip deps, then usually they are coming from a `pyproject.toml`, so it makes sense to do all the configuration there, right?

tlambert03:
yeah, totally fine with me to leave out the cli option

tlambert03:
seem to be some test failures... but I haven't been able to reproduce them locally (macos)

maresb:
This looks pretty good.

I'm somewhat horrified by the way that `lookup.py` currently works. I cleaned up a few of the worst things a while back, but didn't go nearly far enough. If you feel similarly, then please have at it! :grin: (Stuff like trying to get rid of `LOOKUP_OBJECT`, adding an `__init__()` to `_LookupLoader`, etc.)

A lot of this feels a bit overly-fancy to me like `ChainMap` and `Mapping`. I personally prefer working with dicts, so I'd instead do something like
```python
return {**self.local_mappings, **self.remote_mappings}
```
I think it's just less mental effort to understand concrete types. I won't request you to change it, but in case you do make further changes, it'd be nice to go towards that concrete/simple direction.

tlambert03:
> I think it's just less mental effort to understand concrete types. I won't request you to change it, but in case you do make further changes, it'd be nice to go towards that concrete/simple direction.

happy to change it as you prefer!  will have a look at other `lookup.py` cleanup too

maresb:
Haha, I think at least some of the CI failures may be a timing issue. Look what just dropped:

https://github.com/pandas-dev/pandas/releases/tag/v2.1.3

There's no corresponding conda-forge package yet. Maybe some of the tests need pins.

tlambert03:
ha!

tlambert03:
made some slight adjustments to the module.  got rid of the module-level global and removed a duplicated function in `pyproject_toml.py`.  I could probably go father and remove the `_LookupLoader` altogether, using lru_cache on a function call rather than cached_property on the class.  but didn't want to be overly aggressive in my first PR :joy:

conda_lock/lookup.py
```diff
@@ -69,61 +88,49 @@ def local_mappings(self, mappings: Mapping[str, Union[str, MappingEntry]]) -> No
...
-    def pypi_lookup(self) -> Mapping[NormalizedName, MappingEntry]:
-        """ChainMap of PyPI to conda name mappings.
+    @cached_property
+    def pypi_lookup(self) -> Dict[NormalizedName, MappingEntry]:
+        """Dict of PyPI to conda name mappings.
 
         Local mappings take precedence over remote mappings fetched from `_mapping_url`.
         """
-        return ChainMap(self.local_mappings, self.remote_mappings)
+        return {**self.remote_mappings, **self.local_mappings}
```
maresb, line 101: Ah good, you caught that I typed my comment backwards. :joy:

conda_lock/src_parser/pyproject_toml.py
```diff
@@ -422,16 +405,15 @@ def parse_python_requirement(
...
     parsed_req = parse_requirement_specifier(requirement)
-    name = canonicalize_pypi_name(parsed_req.name)
     collapsed_version = str(parsed_req.specifier)
     conda_version = poetry_version_to_conda_version(collapsed_version)
     if conda_version:
         conda_version = ",".join(sorted(conda_version.split(",")))
 
     if normalize_name:
-        conda_dep_name = normalize_pypi_name(name)
+        conda_dep_name = pypi_name_to_conda_name(parsed_req.name)
```
maresb, line 414: Great improvement on the name!!!

mariusvniekerk:
Might be worth adding in a test to ensure that this functionality doesn't regress

tlambert03:
> Might be worth adding in a test to ensure that this functionality doesn't regress

added, thanks!

maresb:
Thanks @tlambert03 for adding the tests! Unfortunately they don't seem to be passing yet.

tlambert03:
thanks. yeah I saw that ... i think there's a fixture/state leakage somewhere (perhaps in the resolver).  It works in isolation.  need to figure out what other test it's conflicting iwth

maresb:
Ah, ya, no pressure. It's always such a pain when tests run locally but not in CI. Good luck and thanks for all your work!

tlambert03:
figured it out.  hopefully should pass now 🤞

maresb (changes requested):
Thanks @tlambert03!

I am proposing some cleanups here: https://github.com/tlambert03/conda-lock/pull/1

Also, I'd prefer to leave out the `MappingEntry` class until the next PR when we actually implement it. (I don't like having the unused code laying around.)

Then I'd like to take one more pass at this since I am still getting confused by the (pre-existing) indirection.

tlambert03:
> Also, I'd prefer to leave out the MappingEntry class until the next PR when we actually implement it. (I don't like having the unused code laying around.)

not sure I'm following this comment.  That class is already there.  Are you proposing we remove it and re-add it later?

maresb:
> not sure I'm following this comment. That class is already there. Are you proposing we remove it and re-add it later?

Ah, my bad, I was confused. I thought you added it since it matched with one of your initial proposals. Also this was reinforced by observing that it seemed to be extraneous. But now I remember that it was originally the schema for an item in the mapping.

As a sort of documentation to make this clear, it would be nice to replace

```python
lookup = {canonicalize_name(k): v for k, v in lookup.items()}
for v in lookup.values():
    v["pypi_name"] = canonicalize_name(v["pypi_name"])
return lookup
```

with something like:

```python
result: Dict[NormalizedName, MappingEntry] = {}
for k, v in lookup.items():
    assert k == v["pypi_name"]
    pypi_name = canonicalize_name(k)
    assert pypi_name == k
    conda_name = v["conda_name"]
    assert pypi_name not in result
    result[pypi_name] = conda_name
return result
```

> The main question is how you want to handle the reverse lookup. If a pypi name can be converted to multiple conda packages, that makes the mapping non invertible

Good question. The whole mapping story is very messy. But currently there is only instance of this happening:

```python
from collections import defaultdict

from conda_lock.lookup import _lookup_loader

pypi_names = defaultdict(set)
for k, v in _lookup_loader.remote_mappings.items():
    pypi_names[v["conda_name"]].add(k)
print({k: v for k, v in pypi_names.items() if len(v) > 1})
# {'pyqt': {'pyqt4', 'pyqt5'}}
```

My suggestion would be to emit a warning and then take a guess. The user could always define a local mapping to break the tie, if local mappings were given priority. :wink:

anibali:
Is there anything I can do to help this along? I would really appreciate the ability to disable the default mappings and just manually specify them from pyproject.toml (I find this a lot more convenient than the recent solution of having a separate mapping file provided via a command line argument). I was going to write my own PR, but stumbled across this in my preliminary search.

My initial thought was something along the lines of:

```toml
[tool.conda-lock.dependencies]
pytorch-scatter = {pypi_name = "torch-scatter"}
```

It would also offer a neat way of being explicit about when the PyPI version of a package is preferred. For example, say I want to install opencv-python from PyPI instead of the conda-forge package:

```toml
[tool.conda-lock.dependencies]
opencv = {source = "pypi", pypi_name = "opencv-python"}
```

I really enjoy using conda-lock, and would love to see the PyPI <-> Conda mapping handling improved by removing the need for another external file.

tlambert03:
i think the core functionality of adding `[tool.conda-lock.pypi-to-conda-name]` was all set to go, then we got sidetracked on cleaning up the existing module patterns.  I'll resolve conflicts here so it can be merged, but probably don't have time to keep cleaning up the previous patterns.

tlambert03:
conflicts resolved.  @maresb, looks like the conflicts, which resulted from #588, were actually already fixed in here.  So, would it be ok with you to merge this and work on cleaning up the legacy lookup.py in some other PR?  i think this is getting caught up in scope creep.

maresb:
@tlambert03, sorry for letting this slip. Do you understand why the tests are failing though?

PR #557, opened 2023-11-20
alphasort env files

### LINKED ISSUE #554: Lock file for `kind=env` should be alphabetized
### Checklist

### What is the idea?

Like https://github.com/conda/conda-lock/issues/491 and https://github.com/conda/conda-lock/pull/524, the `kind=env` lock files should also have the entries sorted. This is required to match the output produced by `conda env export`.

### Why is this needed?

This enables diff of outputs with `conda env export`. As well as matching the behaviour of the v2 lock file and producing a stable order.

### What should happen?

_No response_

### Additional Context

_No response_

jamesmyatt:
Fixes #554

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -608,7 +608,11 @@ def render_lockfile_for_platform(  # noqa: C901
     pip_deps: List[LockedDependency] = []
 
     # ensure consistent ordering of generated file
-    lockfile.toposort_inplace()
+    # topographic for explicit files and alphabetical otherwise (see gh #554)
+    if kind == "explicit":
+        lockfile.toposort_inplace()
+    else:
+        lockfile.alphasort_inplace()
     lockfile.filter_virtual_packages_inplace()
 
     for p in lockfile.package:


jamesmyatt:
There's already a test that the explicit files are toposorted: https://github.com/conda/conda-lock/blob/5bce4228cbd0af96ae00ec8efe0b9f2eb694ab53/tests/test_conda_lock.py#L995. Does it need a test that the env files are alphasorted?

Also should this use the `KIND_...` constants instead of literals?

maresb:
If we want that the output just happens to be alphabetically sorted, then I'd say we don't need a test. But if we declare that alpha-sorting is an actual feature, then we should add a test to ensure it doesn't regress.

> Also should this use the `KIND_`... constants instead of literals?

These constants date to before my involvement with conda-lock. I'm not familiar with them. I'd only recommend using them if they're helpful.

> Also I think the `Union[Literal[...], ...]` type hints can be simplified to `Literal[..., ...]`. Is that right?

That is correct. I am always grateful for such cleanups. We could also remove them if they're not serving a useful purpose.

jamesmyatt:
I don't see any documentation or tests in the conda repo on the ordering of the exported env, so there's no guarantee that won't change either. So we can probably claim that it's an ordering rather than a feature.

I'll leave the constants and the type hints alone. Both could do with bigger PRs. I think rather than 1 method at a time.

jamesmyatt:
Seems to work on my machine 👍

maresb:
@jamesmyatt, in order to address a likely bug I made #560 which will cause a merge conflict. Resolving it will be easy: call `lockfile.filter_virtual_packages_inplace()` after sorting.

Before the next release I need to address #559. If you feel especially motivated, I'd be grateful for support there, but otherwise I'll get to it eventually.

jamesmyatt:
Thanks @maresb. I'll rebase this branch on [maresb:dont-filter-in-toposort](https://github.com/maresb/conda-lock/tree/dont-filter-in-toposort) and mark it as draft then.

maresb:
Hi @jamesmyatt, sorry about the delay on #560. That's now been merged, so this will be ready to merge after you get a chance to rebase. Thanks for your patience!

jamesmyatt:
I've rebased now. Thanks

maresb (approved):

Issue #556, opened 2023-11-20
Invalid/Temporary local packages unexpectedly appearing in conda-lock.yml


zaneselvans:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

After updating to `conda-lock` v2.5.0 some odd new packages are showing up in our `conda-lock.yml` when it is regenerated via a GitHub action:

```yaml
package:
  - name: __glibc
    version: "2.17"
    manager: conda
    platform: linux-64
    dependencies: {}
    url: file:///home/runner/work/_temp/tmp8q60deq6/linux-64/__glibc-2.17-0.tar.bz2
    hash:
      md5: ""
    category: main
    optional: false
  - name: __osx
    version: "11.0"
    manager: conda
    platform: osx-64
    dependencies: {}
    url: file:///home/runner/work/_temp/tmp8q60deq6/osx-64/__osx-11.0-0.tar.bz2
    hash:
      md5: ""
    category: main
    optional: false
  - name: __osx
    version: "11.0"
    manager: conda
    platform: osx-arm64
    dependencies: {}
    url: file:///home/runner/work/_temp/tmp8q60deq6/osx-arm64/__osx-11.0-0.tar.bz2
    hash:
      md5: ""
    category: main
    optional: false
```

I noticed that they exist because the lack of a `hash` value is causing the `setup-micromamba` GitHub action to fail with the following error:

```
  Empty environment created at prefix: /home/runner/micromamba/envs/pudl-dev
  critical libmamba either package 'package.info.hash.md5' or 'package.info.hash.sha256' must be specified, found none
  Error: The process '/home/runner/micromamba-bin/micromamba' failed with exit code 1
```

However, no packages named `__osx` or `__glibc` appear in the `conda-lock.yml` when I generate it locally using a similar command.  The only obvious differences between the two environments are:
- The GitHub runners are running Linux, and my local environment is on an M1 Macbook Pro
- On the GitHub runner it is using `conda-lock --micromamba` while locally I'm using `conda-lock --mamba`. The command being used to generate the conda lockfiles / environment files:

```bash
conda-lock \
	--${mamba} \
	--file=pyproject.toml \
	--lockfile=environments/conda-lock.yml
```

However, running the above commands locally using `--micromamba` does not result in the `__glibc` or `__osx` packages appearing in `conda-lock.yml`

### Conda Info

```shell
active environment : None
            shell level : 0
[...]

PR #560, opened 2023-11-24
Don't filter in toposort

maresb:
### Description

Before this PR, toposort would drop items (the virtual packages). I wouldn't expect a sort operation to do this. This is a major bug hazard, see #556 and also potentially #557. I've spit this out filtering into a separate step.

Other commits implement some minor cleanups along the way.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -609,6 +609,7 @@ def render_lockfile_for_platform(  # noqa: C901
 
     # ensure consistent ordering of generated file
     lockfile.toposort_inplace()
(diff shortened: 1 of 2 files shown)


jamesmyatt:
Any progress on this?

maresb:
Sorry, this stalled out as I couldn't figure out if my assertion was correct.

I suppose we could remove it and proceed blindly, or better replace it with a warning and ask the user to file a report if it's false.

@mariusvniekerk, any insight here? It would be nice to either get test coverage on L122 or remove it as unnecessary.

maresb:
I'm dropping my assertion and will follow it up later.

PR #564, opened 2023-11-30
Fix case when using multiple platforms but pip contains a single platform

basnijholt:
Prevents
```
  File "/Users/bas.nijholt/micromamba/lib/python3.11/site-packages/conda_lock/src_parser/environment_yaml.py", line 58, in _parse_environment_file_for_platform
    for spec in mapping_spec["pip"]:
TypeError: 'NoneType' object is not iterable
```
Which occurs when locking:
```
name: example
channels:
  - conda-forge
dependencies:
  - tomli
  - pip:
    - psutil  # [linux64]
platforms:
  - linux-64
  - osx-arm64
```
Which becomes:
```
env_yaml_data = {'name': 'test-pip-with-platform-selector', 'channels': ['conda-forge'], 'dependencies': ['tomli', {'pip': None}], 'platforms': ['linux-64', 'osx-arm64']}
```
For `osx-arm64`.

This fix skips `{"pip": None}`.


maresb:
Thanks @basnijholt!!!

PR #567, opened 2023-12-08
Fix an issue when specifying full wheel URLs for a pypi package.

romain-intel:
The pypi solver expects hashes to be provided as part of the URL but, in most cases, if a user provides a URL for a package, they will not provide one with a hash. This PR ignores the missing hash instead of crashing.

A test was added that fails prior to this change and succeeds after.

### DIFF
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -214,8 +214,11 @@ def get_requirements(
 
             if op.package.source_type == "url":
                 url, fragment = urldefrag(op.package.source_url)
-                hash_type, hash = fragment.split("=")
-                hash = HashModel(**{hash_type: hash})
+                hash_splits = fragment.split("=")
+                if len(hash_splits) == 2:
+                    hash = HashModel(**{hash_splits[0]: hash_splits[1]})
+                else:
+                    hash = HashModel()
                 source = DependencySource(type="url", url=op.package.source_url)
             elif op.package.source_type == "git":
                 url = f"{op.package.source_type}+{op.package.source_url}@{op.package.source_resolved_reference}"
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -215,10 +215,12 @@ def get_requirements(
             if op.package.source_type == "url":
                 url, fragment = urldefrag(op.package.source_url)
                 hash_splits = fragment.split("=")
-                if len(hash_splits) == 2:
+                if fragment == "":
+                    hash = HashModel()
+                elif len(hash_splits) == 2:
                     hash = HashModel(**{hash_splits[0]: hash_splits[1]})
                 else:
-                    hash = HashModel()
+                    raise ValueError(f"Don't know what to do with {fragment}")
                 source = DependencySource(type="url", url=op.package.source_url)
             elif op.package.source_type == "git":
                 url = f"{op.package.source_type}+{op.package.source_url}@{op.package.source_resolved_reference}"
--- a/tests/test-pip-full-url/environment.yml
+++ b/tests/test-pip-full-url/environment.yml
@@ -6,3 +6,4 @@ dependencies:
   - pip
   - pip:
     - requests @ https://github.com/psf/requests/releases/download/v2.31.0/requests-2.31.0-py3-none-any.whl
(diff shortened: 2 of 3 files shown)


romain-intel:
I am not sure why that one test failed. Feels like more of a fluke and unrelated to the PR.

conda_lock/pypi_solver.py
```diff
@@ -214,8 +214,11 @@ def get_requirements(
 
             if op.package.source_type == "url":
                 url, fragment = urldefrag(op.package.source_url)
-                hash_type, hash = fragment.split("=")
-                hash = HashModel(**{hash_type: hash})
+                hash_splits = fragment.split("=")
```
maresb, line 217: Oh, it looks like we [don't have coverage](https://app.codecov.io/gh/conda/conda-lock/blob/main/conda_lock%2Fpypi_solver.py#L217) for this case. :cry: Let's fix that...

conda_lock/pypi_solver.py
```diff
@@ -214,8 +214,11 @@ def get_requirements(
...
+                if len(hash_splits) == 2:
+                    hash = HashModel(**{hash_splits[0]: hash_splits[1]})
+                else:
+                    hash = HashModel()
```
maresb, line 221: Let's make sure all the cases are thoroughly covered:

```suggestion
                if fragment == "":
                    hash = HashModel()
                elif len(hash_splits) == 2:
                    hash = HashModel(**{hash_splits[0]: hash_splits[1]})
                else:
                    raise ValueError(f"Don't know what to do with {fragment}")
```

tests/test-pip-full-url/environment.yml
```diff
@@ -0,0 +1,8 @@
+# environment.yml
+channels:
+  - conda-forge
+
+dependencies:
+  - pip
+  - pip:
+    - requests @ https://github.com/psf/requests/releases/download/v2.31.0/requests-2.31.0-py3-none-any.whl
```
maresb, line 8: Add a second URL so we get coverage for handling the hash

```suggestion
    - requests @ https://github.com/psf/requests/releases/download/v2.31.0/requests-2.31.0-py3-none-any.whl
    - typing-extensions @ https://files.pythonhosted.org/packages/24/21/7d397a4b7934ff4028987914ac1044d3b7d52712f30e2ac7a2ae5bc86dd0/typing_extensions-4.8.0-py3-none-any.whl#sha256=8f92fc8806f9a6b641eaa5318da32b44d401efaac0f6678c9bc448ba3605faa0
```

tests/test_conda_lock.py
```diff
@@ -2460,3 +2460,25 @@ def test_parse_environment_file_with_pip_and_platform_selector():
...
+        platforms=["linux-64"],
+    )
+
+    lockfile = parse_conda_lock_file(env_file.parent / DEFAULT_LOCKFILE_NAME)
+
+    (requests_dep,) = [p for p in lockfile.package if p.name == "requests"]
+    assert (
+        requests_dep.url
+        == "https://github.com/psf/requests/releases/download/v2.31.0/requests-2.31.0-py3-none-any.whl"
+    )
```
maresb, line 2484:
```suggestion
    (requests_dep,) = [p for p in lockfile.package if p.name == "requests"]
    (typing_extensions_dep,) = [p for p in lockfile.package if p.name == "typing-extensions"]
    assert (
        requests_dep.url
        == "https://github.com/psf/requests/releases/download/v2.31.0/requests-2.31.0-py3-none-any.whl"
    )
    assert requests_dep.hash.sha256 is None
    assert (
        typing_extensions_dep.url
        == "https://files.pythonhosted.org/packages/24/21/7d397a4b7934ff4028987914ac1044d3b7d52712f30e2ac7a2ae5bc86dd0/typing_extensions-4.8.0-py3-none-any.whl"
    )
    assert typing_extensions_dep.hash.sha256 == "8f92fc8806f9a6b641eaa5318da32b44d401efaac0f6678c9bc448ba3605faa0"
```

maresb (changes requested):

maresb:
> I am not sure why that one test failed. Feels like more of a fluke and unrelated to the PR.

Indeed, I suspect that there's an out-of-memory condition or similar that occasionally causes a cascade of failures like that. Upon rerunning it's passing again.

romain-intel:
Thanks for the super quick review. Accepted all changes.

maresb (approved):

PR #582, opened 2024-01-16
Replace unnecessary Popen with subprocess.run

maresb:
#581 raised the issue of I/O deadlock with our use of `subprocess.Popen` and proposed solving this with threads. In this PR I take a step back and suggest that `subprocess.Popen` is unnecessary and can be replaced with `subprocess.run`. This simplifies the subprocess call and obviates the need for threads.


maresb:
Rejected in favor of #581

PR #586, opened 2024-01-18
Add explanation and minor improvement to #581

tadeu:
### Description

Adding an explanation and a suggested improvement to #581 - it seems important to make it explicit why the thread is there.

### DIFF
--- a/conda_lock/invoke_conda.py
+++ b/conda_lock/invoke_conda.py
@@ -116,6 +116,11 @@ def _invoke_conda(
         encoding="utf-8",
     ) as p:
         stdout = []
+        # Using a thread so that both stdout and stderr can be consumed concurrently.
+        # This avoids a potential deadlock when the child conda process is trying to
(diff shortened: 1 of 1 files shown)


tadeu:
Sorry for not doing this together with the original PR before :blush:

maresb (approved):

maresb:
Thanks a lot for the comment, I find it very helpful.

Does this use of `stdout_thread` make any functional difference? Like if `p.stdout` is true once is it always true?

tadeu:
> Does this use of `stdout_thread` make any functional difference? Like if `p.stdout` is true once is it always true?

No functional difference, just maintainability: in case the code is refactored and `if p.stdout:` is changed to something different, the other `if p.stdout:` would also need to be changed (and it's a bit far away) - now there's no way to change one condition and "forget" to change the other one.

nicoddemus (approved):

Issue #579, opened 2024-01-09
"pypi_to_conda_lookup_file" flag not working as intended


neojunjie:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

The occurrence of an error was observed when attempting to reference an alternative location for the mapping file using the "pypi_to_conda_lookup_file" option.

Command fired: 
```bash
conda-lock -f test_conda_lock.yml -p linux-64   --pypi_to_conda_lookup_file https://raw.githubusercontent.com/regro/cf-graph-countyfair/master/mappings/pypi/grayskull_pypi_mapping.yaml
```

Error Logs:
```python
Traceback (most recent call last):
  File "/home//miniforge3/bin/conda-lock", line 8, in <module>
    sys.exit(main())
  File "/home//miniforge3/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home//miniforge3/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home//miniforge3/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home//miniforge3/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home//miniforge3/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home//miniforge3/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home//miniforge3/lib/python3.10/site-packages/conda_lock/conda_lock.py", line 1319, in lock
    set_lookup_location(pypi_to_conda_lookup_file)
  File "/home//miniforge3/lib/python3.10/site-packages/conda_lock/lookup.py", line 66, in set_lookup_location
    LOOKUP_OBJECT.mapping_url = lookup_url
  File "/home//miniforge3/lib/python3.10/site-packages/conda_lock/lookup.py", line 27, in mapping_url
    del self.pypi_lookup
AttributeError: pypi_lookup
```

maresb:
Thanks for the report! This was fixed in #588 and has been released.

PR #588, opened 2024-01-24
Proposed fix for #579

ianpye:
### Description

This PR proposes a simple fix to issue #579. Currently, the --pypi_to_conda_lookup_file flag does not work and produces an error whenever it is used, without checking the link first. 

The removal of the following lines allows the flag to work as intended:

```
del self.pypi_lookup
del self.conda_lookup
```


maresb:
Thanks a lot @ianpye for taking the initiative on this!

I believe the proper fix is something more like:

```python
        if self._mapping_url != value:
            self._mapping_url = value
            # Invalidate cache
            try:
                del self.pypi_lookup
            except AttributeError:
                pass
            try:
                del self.conda_lookup
            except AttributeError:
                pass
```

This feature desperately needs a test to prevent a similar regression in the future.

maresb:
Thanks @ianpye!

I made a quick attempt to write a test, but I didn't manage to get it to work properly yet. Do you have any ideas for a brief test that ensures we are using the specified mapping? I was thinking we could create a mapping URL that [swaps some dependency](https://gist.githubusercontent.com/maresb/a019fd575617b23f94f450be87288f39/raw/24989171d577218b9e5bc2003bb9a8ab12eab81e/gistfile1.txt), and then the test checks inside the generated lockfile for that substituted dependency.

maresb:
Maybe I'm going the wrong way in my gist. I was thinking about constructing an example for `pyproject.toml`, but you're using only `environment.yml`, which is probably simpler.

maresb:
The current test failure on macos is erroneous. See #591.

PR #590, opened 2024-01-29
Add flags to disable pypi.org and add own private pip repository links

ianpye:
Currently, when creating a conda-lock file using `environment.yml`, poetry will attempt to make a request to `pypi.org`, with the user being unable to disable it, resulting in a conda-lock exception when used in an environment that does not allow access to `pypi.org`. It is only currently possible to make such configurations when creating lock files using `pyproject.toml`.

With these 2 CLI flags, we will be able to both separately disable requests to pypi.org and add our own private mirror links.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -270,6 +270,8 @@ def make_lock_files(  # noqa: C901
     metadata_yamls: Sequence[pathlib.Path] = (),
     with_cuda: Optional[str] = None,
     strip_auth: bool = False,
+    disable_pypi_requests: bool = False,
+    private_pip_repositories: Optional[Sequence[str]] = None,
 ) -> None:
     """
     Generate a lock file from the src files provided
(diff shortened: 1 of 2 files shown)


maresb:
Adding a CLI flag to disable PyPI makes a lot of sense to me, and seems straightforward.

Are you able to add private repos from `environment.yml` like [this example](https://github.com/conda/conda-lock#private-pip-repositories)? I'm hesitant to add another a CLI method to specify private repos, in particular because the environment variable substitution for credentials would be messy. (It'll be difficult to know if the substitution is being done by Bash or conda-lock.)

Simard302:
I am very happy to see this PR. My server can not consistently connect to `pypi.org` and I need to disable it to use mirrors instead. For now, I had to avoid using conda-lock entirely until this is resolved.

@ianpye @maresb what is the status of this PR. If any help is needed, I am happy to contribute.

maresb:
I apologize for not finding more time to dedicate to conda-lock and to see through features like this.

Glancing over this, it seems very straightforward. I don't think this is problematic. If either of you @ianpye or @Simard302 could rebase this I think I could merge and release it, although it could take some days for me to respond.

In general if you're dealing with mixed Conda and pip dependencies I think you'd be better served with [pixi](https://docs.databricks.com/en/partners/bi/power-bi.html). Their support for mixed environments is quite a bit more stable. But I'm not sure how far along they are with private repo support.

Issue #596, opened 2024-02-05
conda-lock fails with micromamba and ensureconda 1.4.4


HedgehogCode:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

If I run
```
$ conda-lock --micromamba --no-mamba -f environment.yml -p linux-64
```

I get the error
```
Traceback (most recent call last):
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/bin/conda-lock", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benjamin/miniconda3/envs/tmp-conda-lock/lib/python3.12/site-packages/conda_lock/conda_lock.py", line 1398, in lock
    lock_func(
[...]

PR #597, opened 2024-02-05
Increase minimum ensureconda version for py3.12 compatibility

jfrost-mo:
Prior to ensureconda v1.4.4 distutils is used, which was removed in python 3.12. This prevents conda-lock running under python newer than 3.11, so the minimum version of ensureconda is increased.

Fixes https://github.com/conda/conda-lock/issues/542
Fixes https://github.com/conda/conda-lock/issues/596

The fix involves switching from distutils.version to packaging.version from the [packaging package](https://pypi.org/project/packaging/), which is maintained by the pypa.

The conda-lock.yml files will need regenerating after this PR.

PR #598, opened 2024-02-05
Replace distutils.LooseVersion with packaging.Version

jfrost-mo:
distutils was removed in python 3.12, so this is needed for conda-lock to work beyond python 3.11.

Fixes https://github.com/conda/conda-lock/issues/596

Will require the conda-lock.yml files regenerating due to the updated dependency.


maresb:
Thanks for this! I'm glad that our changes were identical. Let's continue in #597.

PR #616, opened 2024-03-08
Update header cmdlines in generated lockfiles (#518)

### LINKED ISSUE #518: Faulty code snipped in lock file header
### Checklist

### What happened?

I generate a lock file using v2.3 and in the header the follow code snipped occurs.

```
#
#
...
```

So now I try to install it into a new env with 

```
conda-lock install -n myenv --file conda-lock.yml
```

But then I get the following error

```
Usage: conda-lock install [OPTIONS] [LOCK_FILE]
Try 'conda-lock install --help' for help.

Error: No such option: --file Did you mean --auth-file?
```

### Conda Info

```shell
-
```

### Conda Config

```shell
-
```

### Conda list

```shell
-
```

### Additional Context

-

timsnyder:
Command lines suggested for install of the generated lockfile, which are put in the header of the generated lockfile, incorrectly suggest the use of a `--file` argument.

`--file` is no longer an option in the conda-lock CLI.  The lockfile is a positional parameter.

fixes #518

### DIFF
--- a/conda_lock/lockfile/__init__.py
+++ b/conda_lock/lockfile/__init__.py
@@ -187,14 +187,14 @@ def write_conda_lock_file(
                 available, unless you explicitly update the lock file.
 
                 Install this environment as "YOURENV" with:
-                    conda-lock install -n YOURENV --file {path.name}
+                    conda-lock install -n YOURENV {path.name}
                 """
             )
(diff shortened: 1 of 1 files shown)


maresb:
Thanks @timsnyder!

PR #628, opened 2024-04-25
Fix mypy precheck failures

benmoss:
### Description

The precommit hook is failing in https://github.com/conda/conda-lock/pull/626 (among other places) with these errors:

```
conda_lock/invoke_conda.py:197: error: Non-overlapping equality check (left operand type: "Channel", right operand type: "Literal['defaults']")  [comparison-overlap]
conda_lock/conda_solver.py:349: error: Non-overlapping equality check (left operand type: "Channel", right operand type: "Literal['defaults']")  [comparison-overlap]
```

I'm not a Python expert but I think the problem is we're comparing a `Channel` with a string, and equality isn't implemented for those two types. We could either just compare the `url` field or turn the `"defaults"` string into a channel, not sure which is preferred.

### DIFF
--- a/conda_lock/conda_solver.py
+++ b/conda_lock/conda_solver.py
@@ -2,7 +2,6 @@ import json
 import logging
 import os
 import pathlib
-import shlex
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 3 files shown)


maresb:
Thanks a lot @benmoss for taking initiative on this!!! (I had a very busy week.)

The msys2 stuff predates my involvement in conda-lock, so I needed to do a bit of research to understand what it's all about. I've added a test case to ensure that the newly-activated code is actually correct.

maresb:
We should be able to cut a new release as soon as everything is confirmed green.

PR #636, opened 2024-05-07
[WIP] Update to poetry 1.8.2 (minimal changes)

romain-intel:
### Description

This addresses #613 and updates Poetry to a more modern release (in particular taking in changes around support for Metadata from pypi indices).

This is still a WIP, still missing:
  - proper setup for the environment (needs support for environment markers to properly resolve cross platform)
  - fix the failing tests (hash override) -- I am not sure if this is still relevant as well so not 100% sure how to proceed here.
  - verification of the small code changes that were done to support the new APIs
Putting it out here so people can take a look and see if there is anything that should be addressed first.

Issue #613, opened 2024-02-26
Update vendored Poetry to 1.8.0


romain-intel:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Poetry 1.8.0 came out (see here: https://github.com/python-poetry/poetry/releases/tag/1.8.0) and it adds some significant improvements particularly around resolution speed (support for PEP 658 as well as range requests for repos that do not support PEP 658) as well as correctness issues (https://github.com/python-poetry/poetry/pull/9000 -- an issue that bit me more than once).

It would be nice to move from poetry 1.1.15 (the one currently vendored) to this one.

### Why is this needed?

Apart from the points raised above (speed and correctness in certain situations), the version of poetry currently vendored is quite old and upgrading it would be a good idea to keep up with newer developments. Poetry now officially supports python 3.12 which would be a side benefit.

### What should happen?

Nothing would change for the user except that resolving environments containing pypi and conda dependencies would probably be faster and less expensive (in network cost). More environments would probably be able to work properly as well.

### Additional Context

I am happy to provide help/do this but any guidance would be helpful. I saw some scripts/README around upgrading poetry but I am not sure if there is any additional context around it. I can open a PR for it if this is something that would be accepted (ie: I don't necessarily want to spend a ton of time if it won't be looked at/reviewed and has no path to acceptance :)).

romain-intel:
Waiting a tad -- poetry 1.8.0 seems to have some issues being worked our around their support for PEP 658 (lazy-wheel).

maresb:
Amazing, thanks so much @romain-intel !!!!

romain-intel:
I added an initial PR. It's not done yet but hopefully gives you a scope of the work (it wasn't too bad).

PR #629, opened 2024-04-26
Add support for pip hash checking

sfinkens:
### Description

Fix #624 by

1. Adding support for pip's `--hash` option in the requirement parser, for example 
```
- pip:
    - my_module === 1.2.3 --hash=sha256:1234
```
2. Using that hash to create a lockfile

### Questions

Is this going into the right direction? And some more specific questions:

1. The unit test is mostly copied from other another test. Should this be factorized?
2. This solution only works for a single hash. The [documentation](https://pip.pypa.io/en/stable/topics/secure-installs/#multiple-hashes-per-package) states that you can specify multiple hashes for the same package, for example for different platforms. As far as I can see this is not supported by conda-lock's hash model. But maybe it's not needed, because conda-lock supports different platforms anyway.
3. I created two new subclasses which hold the hash, mostly because I didn't want to break anything. But I'm not sure if that's the best approach.
4. I'm not sure how to resolve the mypy issue in `pypi_solver.py` (`PoetryDependency` not having the `get_hash_model` method). I agree with mypy that the solution is not really consistent, because some dependencies still carry the hash in their URL. Maybe `PoetryURLDependency` and `PoetryVCSDependency` could be extended with a `get_hash_model` method as well?

PR #637, opened 2024-05-07
Update Poetry to 1.8.3

romain-intel:
### Description

Replaces #636 but refactors commits to make it easier to review. Addresses #613 

Still todo:
- proper setup for the environment (needs support for environment markers to properly resolve cross platform)
- fix the failing tests (hash override) -- I am not sure if this is still relevant as well so not 100% sure how to proceed here.
- verification of the small code changes that were done to support the new APIs


maresb:
Thanks so much, the commit structure looks great!

romain-intel:
I know it's not done fully yet but let me know what you think particularly about the failing test (if I should work to fix it or if the functionality is no longer needed). I am also working on adding markers to properly be able to resolve the environment (so new feature but I guess needed to finish this).

maresb:
Hi @romain-intel, it looks great so far!

The failing test was recently added in #629 so it'd be great to fix that to ensure no regression. Perhaps @sfinkens could help?

sfinkens:
The problem seems to be that `_vendor.poetry.core.packages.package.Package.to_dependency()` converts `PoetryDependencyWithHash` to an "ordinary" `Dependency`. So the hash from the environment file is lost and a new one is obtained from the online repository.

I don't know if it makes sense to just handle this extra case in `to_dependency` or if there's a more elegant way.

maresb:
Thanks @sfinkens for the clue!

I think I've fixed it in 630adf8. Could you please review?

maresb:
Looks like there's one remaining test failure: `test_it_uses_pip_repositories_with_env_var_substitution[conda-response_url_without_credentials]`. @jacksmith15, is this something you could help with?

sfinkens:
> Thanks @sfinkens for the clue!
> 
> I think I've fixed it in [630adf8](https://github.com/conda/conda-lock/commit/630adf89bf96f96575fcbc737b7a787538b8be6a). Could you please review?

:+1: Very elegant solution!

romain-intel:
Am back from a quick break so let me try to wrap this up. I know I still need to add environment markers (new feature but seems to be needed to deal with the deps for poetry and create the dev environments). Apart from that, are all the tests ok (I see some failures but not sure if they are flaky/expected). Let me know what else you think is needed on this and I'll get it wrapped up. Thanks for fixing the tests!!

maresb:
Thanks so much @romain-intel for all your work on this, and I'm glad you're back!

I'm curious why [environment markers](https://python-poetry.org/docs/dependency-specification/#using-environment-markers) are necessary.

As for what else is needed, I think we'll primarily have to lean on the test suite. Coverage is far from perfect, but hopefully this fixes more bugs than it creates.

romain-intel:
@maresb : it's because of this: `'xattr >=1.0.0,<1.1.0 ; sys_platform == "darwin"',`. This is a requirement for the new poetry I included and since we use conda-lock to resolve the development environment cross platform, I thought I would try to get it to work. Of course, it's not actually functionally required and we could just have different pyproject.toml to use as sources for conda-lock for the different environments if that is also an acceptable solution (definitely a simpler one). I was trying to keep things exactly the same but happy to take other opinions (and tbh, I don't yet fully know how to pass environment markers all the way down and it may be a bit bigger change than I would like for this). If you guys are fine with just having different .toml files as sources to generate the development environments, I don't need environment markers.

For the tests, there are errors that seem to hint at a bad authorization token or something but am not sure it's related to the PR. Maybe I'll try re-running to see if that was flaky so I wasn't sure if there were actual broken tests. I'll check again.

Let me know what you think for the first thing. Happy to do the "simpler" solution.

maresb:
Thanks for the clarification @romain-intel!

To flesh out exactly what the issue is, as I understand it, the current behavior of conda-lock is to ignore [PEP 508 environment markers](https://peps.python.org/pep-0508/). We dogfood by generating lockfiles for conda-lock using conda-lock itself. The `xattr` dependency does not exist on Windows, so ignoring `sys_platform == "darwin"` breaks this locking workflow when the Windows platform is locked.

While embarrassing, I think I could tolerate breakage of the dogfood for now. It would be much nicer if we could implement a very basic version of environment specifiers, e.g. something that _only_ supports expressions of the form `sys_platform == x`, and I'd rather do it at the conda-lock level rather than the Poetry level. In other words I wonder if it would make things easier if we aim for rough parity with our extremely crude [selectors support](https://github.com/conda/conda-lock/blob/main/conda_lock/src_parser/selectors.py).

I agree that it doesn't make much sense, but I believe there is indeed something in the Poetry upgrade that's causing the weird authorization failures, since other PRs are green, e.g. https://github.com/conda/conda-lock/pull/644. I'll ping the author @jacksmith15 once again in the hopes that he's available to take a look at those auth failures.

romain-intel:
Got it, so to summarize:
- ok to break the dogfooding by having different files to generate the envs
- figure out the broken tests. I'll see if I can figure it out as well but any help would be welcome as well :)

maresb:
> by having different files to generate the envs

I suspect there's a miscommunication here. Wouldn't having different envs work around the dogfooding breakage? I'm suggesting to simply break it.

Anyways, feel free to do whatever makes sense to you. I have the highest trust in your judgement.

romain-intel:
That's a dangerous proposition :).

At any rate, it seems the only actual failure is this: `_ test_it_uses_pip_repositories_with_env_var_substitution[conda-response_url_without_credentials] _`. I'll check around and fix it and also update the branch. Am going to try to finish this this week to get it ready :)

romain-intel:
This is really weird. Running that test independently on my mac works fine. Let me keep digging.

maresb:
@romain-intel, thanks so much for your persistence!!!

I believe Poetry's cache was preventing you from being able to reproduce the error locally, so I modified the test to clear the cache. This allowed me to drill down with a debugger. While it was hard to see exactly what was going on, what was clear is that Poetry wants to handle basic auth through the config dictionary, not the URL.

Upon stripping the creds from the URL and passing them instead via the config dict, the tests pass.

Would you (or anyone else tracking this) be able to critically review my changes?

As for an initial self-review:

1. The config dictionary is mutated by the `config.merge` method. I don't understand all the Poetry internals to know for sure if each Poetry repository gets a reference to the same config or if there's a copy operation at some point. I think this won't matter in the end though, as long as each repository knows its own credentials.
2. Along these lines, I don't think we have any tests for multiple different passwords corresponding to multiple pip repositories at the same time. Probably we should be testing this, but it seems a bit out-of-scope here.
3. Due to the need for Poetry's cache to be cleared, since we're using pytest-xdist, ideally use a locking mechanism (e.g. a lockfile... of the other type :joy:) to prevent the multiple parameter configurations of `test_it_uses_pip_repositories_with_env_var_substitution` from executing simultaneously. There are a few suggestions in this direction on [this page](https://pytest-xdist.readthedocs.io/en/latest/how-to.html#making-session-scoped-fixtures-execute-only-once) but there's still some work required in order to achieve what I described, though it doesn't seem difficult.

romain-intel:
Ah ha! The cache. Of course the cache. Thank you thank you thank you. OK, I'll take a look at the code changes and see if I can add what you describe in your self review. Progress!!!

conda_lock/models/pip_repository.py
```diff
@@ -19,11 +32,24 @@ def base_url(self) -> str:
...
+        return UsernamePasswordDict(username=username, password=password)
+
     @property
     def stripped_base_url(self) -> str:
         """The base URL of the pip repository, without any basic auth."""
-        base_url = urlparse(self.base_url)
-        return urlunparse(base_url._replace(netloc=base_url.netloc.split("@", 1)[-1]))
+        return stripped_url(self.base_url)
+
+    @property
```
romain-intel, line 49: nit: stripped_base_url and stripped_url are the same thing. Do we need/want both?

romain-intel (commented):
Changes look nice :)

conda_lock/models/pip_repository.py
maresb: I don't think they are the same. Should have added example values. :see_no_evil:

tests/test_pip_repositories.py
```diff
@@ -132,6 +134,27 @@ def configure_auth(monkeypatch):
...
+    to_delete = DEFAULT_CACHE_DIR.resolve()
+    assert to_delete.name == "pypoetry-conda-lock" or (
+        to_delete.parent.name == "pypoetry-conda-lock" and to_delete.name == "Cache"
+    )
+
+    # Grab a lock
+    with FileLock(to_delete.parent / ".conda_lock.lock"):
+        yield
+        # Do another independent check that triggers even if we're in optimized mode
+        if "pypoetry-conda-lock" in to_delete.parts:
```
maresb, line 154: Oh, there's a risk of silent failure if this condition is false. We should fail hard.

What happens if the lockfile doesn't get cleaned up? Will the tests just silently hang waiting for release?

conda_lock/models/pip_repository.py
romain-intel: oh right, sorry, I stupidely missed base_url vs url in the arg :(

tests/test_pip_repositories.py
romain-intel: Sounds good. I'll fail hard. Let me see if I can get something to clean it up better. I suspect they may hang but maybe what we can do is fail hard too (a timeout there) and I can maybe introduce a general cleanup at the end of the whole test.

romain-intel:
Since I am so slow, also updated to 1.8.3. Hopefully this is it for this :).

maresb:
I don't really have time to track it down at the moment, but frustratingly I can't run pytest locally because I get:
```
conda_lock/_vendor/poetry/installation/wheel_installer.py:10: in <module>
    from installer import install
E   ModuleNotFoundError: No module named 'installer'
```

At least it works on the CI

romain-intel:
Let me look at the changes today and rebase everything to have update to 1.8.3 in a single commit (reordering stuff). Not sure about the installer though :(.

maresb:
Ya, it's weird that they have an absolute import. At first glance this really looks like a bug in Poetry if Poetry weren't controlling their own environment.

maresb:
Oh! :man_facepalming: Installer is a [package](https://pypi.org/project/installer/)! I need to recreate my environment.

romain-intel:
OK, I think I made both changes. It is indeed much cleaner :). Let me know what you think.

maresb:
Incredible @romain-intel! 

I audited f43e218dc59f2d927556f1acc18a157f97b5198e because that's way too much to look at.

I reproduced it exactly with the following commands:

```bash
vendoring sync -vvv .
dos2unix conda_lock/_vendor/poetry/core/_vendor/lark/grammars/*
dos2unix conda_lock/_vendor/poetry/core/_vendor/fastjsonschema/*
dos2unix conda_lock/_vendor/conda/_vendor/boltons/LICENSE
dos2unix conda_lock/_vendor/poetry/core/_vendor/lark/LICENSE
```

I honestly didn't think this was feasible, but incredibly it seems to me like we're done! Thanks so much!!! :heart:

romain-intel:
whoohoo!! Thanks for merging!!!

maresb:
Important reference: the version pins here are incorrect and have been fixed in #678.

PR #646, opened 2024-06-11
Replace logger.warn w/ logger.warning

zklaus:
### Description

This replaces calls to `logger.warn` with `logger.warning`. The former has been deprecated since Python 3.3 and generates `DeprecationWarning`s that can make output particularly in debugging and testing more noisy than necessary.

Also see conda/conda#13963.

### DIFF
--- a/conda_lock/conda_solver.py
+++ b/conda_lock/conda_solver.py
@@ -230,12 +230,12 @@ def _get_repodata_record(
                 with open(record) as f:
                     repodata: FetchAction = json.load(f)
                 return repodata
-        logger.warn(
+        logger.warning(
             f"Failed to find repodata_record.json for {dist_name}. "
             f"Retrying in 0.1 seconds ({retry}/{NUM_RETRIES})"
         )
         time.sleep(0.1)
-    logger.warn(f"Failed to find repodata_record.json for {dist_name}. Giving up.")
+    logger.warning(f"Failed to find repodata_record.json for {dist_name}. Giving up.")
     return None
 
 


maresb:
Thanks @zklaus!

Most of the changes are on vendored code, so that should not be changed outside of revendoring or specific patches.

The changes to `conda_lock/conda_solver.py` look good though! Would you be able to rebase with just this? Thanks!

zklaus:
Thanks for the quick feedback, @maresb! I did put those changes into a separate commit because I expected this discussion :) The changes in vendored code are all in conda. The corresponding PR for conda is conda/conda#13963.

What's the procedure for revendoring here?

maresb:
Roughly the procedure is

1. Install [vendoring](https://pypi.org/project/vendoring)
2. Update [vendor.txt](https://github.com/conda/conda-lock/blob/main/conda_lock/_vendor/vendor.txt) to update the version
3. Run `vendoring update`
4. Fix breakage

Now may actually be a good time to revendor since we can do it concurrently with https://github.com/conda/conda-lock/pull/637. Thanks for your interest in this @zklaus!

zklaus:
Hm. If we are doing this, it may be best to wait for a release of conda that includes the fixes, so how about this:

1. I'll remove the changes to the vendored code here
2. We merge this PR and consider it enough to unblock conda/conda#13963
3. We merge conda/conda#13963
4. Wait for the next conda release
5. Revendor to that version
6. Profit

maresb:
perfect, thanks @zklaus!!!

PR #647, opened 2024-06-13
Correct mistake in basic usage documentation

### LINKED ISSUE #645: Fix documentation regarding relocking
It seems that part of #632 doesn't accurately reflect the current behavior of how existing lockfiles are used when relocking.

https://github.com/conda/conda-lock/pull/632#issuecomment-2156733823

sfinkens:
### Description

When adding new packages to the environment, the previous solution is _not_ used as constraint. Closes #645


maresb:
Thanks a lot @sfinkens!

sfinkens:
Sure, sorry for the confusion

PR #654, opened 2024-06-28
Sort BaseLockedDependency.dependencies when alpha-sorting a Lockfile

peterbygrave:
### Description

When writing a conda-lock.yml file, the `package.dependencies` are not sorted. This can cause more lines to change than strictly needed when e.g. adding a new package that does not alter any existing packages. This help with merge conflicts etc.

I was unsure of the ethos around testing in this repo, but very happy to add tests as needed with some guidance 🙂

### DIFF
--- a/conda_lock/lockfile/v1/models.py
+++ b/conda_lock/lockfile/v1/models.py
@@ -68,7 +68,7 @@ class BaseLockedDependency(StrictModel):
         return v
 
     def alphasort_inplace(self) -> None:
-        self.dependencies = {k: v for k, v in sorted(self.dependencies)}
+        self.dependencies = {k: self.dependencies[k] for k in sorted(self.dependencies)}
 
 
 class LockedDependency(BaseLockedDependency):
--- a/conda_lock/lockfile/v1/models.py
+++ b/conda_lock/lockfile/v1/models.py
@@ -67,6 +67,9 @@ class BaseLockedDependency(StrictModel):
             raise ValueError("conda package hashes must use MD5")
         return v
 
+    def alphasort_inplace(self) -> None:
+        self.dependencies = {k: v for k, v in sorted(self.dependencies)}
+
 
 class LockedDependency(BaseLockedDependency):
     optional: bool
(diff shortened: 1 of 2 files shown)


conda_lock/lockfile/v1/models.py
```diff
@@ -67,6 +67,9 @@ def validate_hash(cls, v: HashModel, values: Dict[str, typing.Any]) -> HashModel
             raise ValueError("conda package hashes must use MD5")
         return v
 
+    def alphasort_inplace(self) -> None:
+        self.dependencies = {k: v for k, v in sorted(self.dependencies)}
```
peterbygrave, line 71:
```suggestion
        self.dependencies = {k: self.dependencies[k] for k in sorted(self.dependencies)}
```

maresb:
Thanks! LGTM.

I don't see any need for tests here because I don't think we need to guarantee this as an interface contract.

Do you know what's leading to the unsorted dependencies? Can you provide a minimal example? In my lockfiles they're sorted. Is it coming from pip dependencies?

peterbygrave:
> Thanks! LGTM.
> 
> I don't see any need for tests here because I don't think we need to guarantee this as an interface contract.
> 
> Do you know what's leading to the unsorted dependencies? Can you provide a minimal example? In my lockfiles they're sorted. Is it coming from pip dependencies?

Interesting question, I don't believe it was pip dependencies but will double check my local example where I saw it and post.

peterbygrave:
Closing and reopening to trigger ci

peterbygrave:
> Thanks! LGTM.
> 
> I don't see any need for tests here because I don't think we need to guarantee this as an interface contract.
> 
> Do you know what's leading to the unsorted dependencies? Can you provide a minimal example? In my lockfiles they're sorted. Is it coming from pip dependencies?

Looking at my example (unfortunately I cannot shared due to private repo) it happens for conda packages but only when adding packages to my source `environment.yaml` They were quite innocent (e.g. `types-networkx`). The `conda-lock.yml` started sorted to then became unsorted but only for some packages e.g:
Screenshot 2024-06-30 at 07 19 15

Wonder if this is only happening when loading from an existing lock file. Tried looking for something obvious, but nothing stood out.

Now ready for review.

maresb:
Thanks @peterbygrave! I made a few very minor tweaks and will squash-merge once the tests pass.

PR #655, opened 2024-06-29
take into account branch, tag or rev on git dependencies

sebhmg:
### Problem

conda-lock does not take into account `rev`, `branch` or `tag` for a git dependency in pyprojet.toml
See reference in Poetry documentation about specifying the git revision: https://python-poetry.org/docs/dependency-specification/#git-dependencies

It only considers the revision when expressed in the url as `url@rev`

### Fix
Checks both the url and the `branch`, `tag` or `rev` field. Raises an exception in case the information is specified both ways with different values.

### DIFF
--- a/conda_lock/src_parser/pyproject_toml.py
+++ b/conda_lock/src_parser/pyproject_toml.py
@@ -282,7 +282,16 @@ def parse_poetry_pyproject_toml(
                 version = poetry_version_spec
 
             if "git" in depattrs and url is not None:
-                url, rev = unpack_git_url(url)
+                url, at_rev = unpack_git_url(url)
+                rev = (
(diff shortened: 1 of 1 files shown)


maresb:
Thanks @sebhmg!

Would you be able to add some tests? You could add them to [this pyproject.toml](https://github.com/conda/conda-lock/blob/79462ee38d77b3ad827aaf4196a93571e1000a34/tests/test-poetry-git/pyproject.toml) and update [this test](https://github.com/conda/conda-lock/blob/79462ee38d77b3ad827aaf4196a93571e1000a34/tests/test_conda_lock.py#L758) accordingly,

sebhmg:
likely not a problem anymore in latest version

PR #657, opened 2024-07-04
Do a strict pkg meta check @ release CD workflow

webknjaz:
### Description

The previously used non-strict mode of `twine check` ignores a subset of problems with the README rendering. It is the best practice to have `--strict`. Running it in normal CI allows catching issues way before the time comes to make a release. It's rather unpleasant to get upload failures at the last moment.


maresb:
Thanks @webknjaz!

PR #688, opened 2024-09-10
Use a quicker example for the channel-inversion test

maresb:
### Description

This test case has an average duration of 141 seconds, compared with the next-highest durations in the 70-90s range for `test_run_lock*`.

AFAICT we just need to check that the lockfile references a package with an out-of-order channel when that channel is specified. The newer example should accomplish this with a much lower overhead.


maresb:
Great to see you around @mariusvniekerk!!! I'm just charging full speed ahead because my default assumption is that you're busy.

Please feel free to chime in on anything/everything. I'm especially interested in your feedback on this one.

PR #690, opened 2024-09-10
Improve loading of PyPI mapping

maresb:
### Description

Add caching, and load with ruamel.yaml which is much faster.

### DIFF
--- a/conda_lock/lookup.py
+++ b/conda_lock/lookup.py
@@ -191,6 +191,5 @@ def cached_download_file(url: str) -> bytes:
         except Timeout:
             logger.warning(
                 f"Failed to acquire lock on {destination_lock}, it is likely "
-                f"being downloaded by another process. Retrying...",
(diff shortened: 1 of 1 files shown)


conda_lock/lookup.py
```diff
@@ -95,3 +107,90 @@ def pypi_name_to_conda_name(name: str) -> str:
...
+                    if "ETag" in res.headers:
+                        destination_etag.write_text(res.headers["ETag"])
+                    else:
+                        logger.warning("No ETag in response headers")
+                logger.debug(f"Downloaded {url} to {destination_mapping}")
+                return destination_mapping.read_bytes()
+
+        except Timeout:
+            logger.warning(
+                f"Failed to acquire lock on {destination_lock}, it is likely "
```
mariusvniekerk, line 193:
```suggestion
                f"Failed to acquire lock on %s, it is likely "
```

PR #712, opened 2024-09-27
Add/improve support for Path dependencies and GIT dependencies

romain-intel:
Specifically:
  - adds subdirectory for GIT dependencies
  - adds the ability to pass "path" dependencies to Poetry

### Description

This simply extends support of the arguments that conda-lock can parse in a pyproject.toml and passes them down to Poetry. Specifically it adds support for local directories and wheels as well as support for #subdirectory in GIT packages.


romain-intel:
Need to add a test for subdirectory.

romain-intel:
I suspect the micromamba test failures may be due to the release of micromamba 2.0

maresb:
> I suspect the micromamba test failures may be due to the release of micromamba 2.0

Noooooo! :scream: I wanted to release! Ok, I guess we have more work.

Regarding this PR, it looks mostly great. One of my big concerns is favoring Poetry-style `pyproject.toml` over standards-based `pyproject.toml`. I think/hope that this would be a really simple tweak to achieve this. I think you basically just need to support

```python
parse_python_requirement("/path/to/some-package", manager="pip", mapping_url=DEFAULT_MAPPING_URL)
```

Looking forward to the additional test for `subdir`. Thanks again @romain-intel!!!

maresb:
Ok, micromamba v2 was not so bad. I believe I took care of it in https://github.com/conda/conda-lock/pull/713. Feel free to rebase on that.

romain-intel:
Hopefully that works -- the PEP for local requirements seems to be `mypkg @ /path/to/mypkg` so that is what I am testing here. Let me know if you need anything else.

maresb:
On first glance this looks great, thanks so much @romain-intel!!!

Note one hiccup in the doctest with a missing `subdir=None`.

romain-intel:
whoops. Sorry about that. Hopefully that will do it.

Unrelated: I see a flurry of activity, planning a release soon?

romain-intel:
Gah -- something up with windows :(

maresb:
Oops, https://github.com/conda/conda-lock/pull/717 should fix it. Could you please rebase again?

maresb:
> Unrelated: I see a flurry of activity, planning a release soon?

Yes, I really want to get this released as soon as possible. There are a few tricky things though. The release of micromamba 2 is still causing a lot of chaos, e.g. https://github.com/conda/conda-lock/pull/709, so we should probably let that settle down a bit first.

romain-intel:
Rebased.

And yes, micromamba v2 is also causing me headaches :). Hopefully you can release soon though. Definitely interested in getting this large change integrated.

maresb (approved):
This looks good to me, are you ready to merge @romain-intel?

maresb:
I'm going to take all the green I can get and merge this. Feel free to do any followup in a subsequent PR.

romain-intel:
Thanks -- yes, sorry, busy day. Thanks for merging!!

maresb:
I had a concern regarding the `parse_python_requirement` implementation for path dependencies, but now I'm convinced it's not that big of a deal.

We're assuming a nice [PEP 508](https://packaging.python.org/en/latest/specifications/dependency-specifiers/) compliant input string where the name is specified. However, pip is unfortunately [a bit more loose](https://pip.pypa.io/en/stable/reference/requirements-file-format/#example) about actually specifying the name.

So if I do:

```python
from conda_lock.lookup import DEFAULT_MAPPING_URL
from conda_lock.src_parser.pyproject_toml import parse_python_requirement

parse_python_requirement(
    "./downloads/numpy-1.9.2-cp34-none-win32.whl", mapping_url=DEFAULT_MAPPING_URL
)
```

then I get
```
InvalidRequirement: Expected package name at the start of dependency specifier
    ./downloads/numpy-1.9.2-cp34-none-win32.whl
    ^
```

That error message is a lot more informative than I was expecting, which is why I don't think this is so serious of a problem.

Since not everyone has recently read PEP 508, it might be kind to intercept this exception and reraise with a note explaining that names can be specified by `package-name @ ...`. But I wouldn't block the release on this.

romain-intel:
I can make a more informative message. I'll try this over the weekend or early next week. And yes, I am aware of that and it is spec compliant if not the easiest. I think pip extracts the name from the setup or something when not specified. So nice but then again, why have specs :).

PR #713, opened 2024-09-28
Support Micromamba v2

maresb:
### Description
Looks like some things broke?

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -80,7 +80,10 @@ jobs:
       - name: Install Conda environment with Micromamba
         uses: mamba-org/setup-micromamba@v1
         with:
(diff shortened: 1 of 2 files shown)


maresb:
Ok, I think the failures are covered by these two issues in micromamba:
https://github.com/mamba-org/mamba/issues/3467
https://github.com/mamba-org/mamba/issues/3480

maresb:
There were two failures:
* One in the normal test suite where we used `micromamba list --json` to verify the dependencies of an environment. That was affected by https://github.com/mamba-org/mamba/issues/3480, but as far as I can tell, this is not used by conda-lock itself, just by the verification.
* In the integration tests when micromamba tries to install pip dependencies on Windows it fails as per https://github.com/mamba-org/mamba/issues/3467. Since this is not our fault I'm ignoring it by pinning micromamba to 1.5.10.

PR #720, opened 2024-10-03
Unpin micromamba<2 in CI

maresb:
### Description

Now that setup-micromamba v2 is out, I'm going to try and revert #713 and #717.

Not everything worked, so splitting the broken stuff into #721.


maresb:
Now blocked on https://github.com/mamba-org/mamba/issues/3576

PR #735, opened 2024-10-18
Improve error message when update fails due to channel conflict

### LINKED ISSUE #734: Unclear error message when adding a new channel
When I add a new channel to my environment.yml (going from just conda-forge to conda-forge + pytorch), I get an error message Image

The fix was to delete the old conda-lock.yml before locking the environment, but this was not clear from the error message.

maresb:
### Description

Closes #734

Thanks @andrewpollard for the report! Indeed the previous error message is not so great. Could you please review what I wrote here? Would this have been more clear for you, or do you have an idea for how to make it better?

### DIFF
--- a/conda_lock/lockfile/v2prelim/models.py
+++ b/conda_lock/lockfile/v2prelim/models.py
@@ -61,9 +61,13 @@ class Lockfile(StrictModel):
         elif not isinstance(other, Lockfile):
             raise TypeError
 
-        assert (
-            self.metadata.channels == other.metadata.channels
-        ), f"channels must match: {self.metadata.channels} != {other.metadata.channels}"
+        if self.metadata.channels != other.metadata.channels:
+            raise ValueError(
+                f"Cannot merge locked dependencies when the channels are not "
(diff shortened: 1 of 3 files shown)


conda_lock/lockfile/v2prelim/models.py
```diff
@@ -61,9 +61,13 @@ def merge(self, other: "Optional[Lockfile]") -> "Lockfile":
         elif not isinstance(other, Lockfile):
             raise TypeError
 
-        assert (
-            self.metadata.channels == other.metadata.channels
-        ), f"channels must match: {self.metadata.channels} != {other.metadata.channels}"
+        if self.metadata.channels != other.metadata.channels:
+            raise ValueError(
+                f"Cannot merge locked dependencies when the channels are not "
+                f"consisitent. {self.metadata.channels} != {other.metadata.channels}. "
```
AndrewPollard, line 67: Typo "consisitent" -> "consistent"

maresb: Good catch, thanks!

Issue #725, opened 2024-10-04
Cut the conda-lock v3.0.0 release


maresb:
I want to let everyone know my thinking regarding the release process of conda-lock v3 and also prepare the release notes. Comments welcome!

My current impression is that the recent release of mamba/micromamba v2 has caused a bit of chaos, and that it's prudent to wait for things to stabilize somewhat before cutting this release. (Ideally this specifically means merging #720 and #724.) On the other hand, I expect this release to resolve many outstanding bugs and blockers, so many will want to install as soon as possible.

I'd like to encourage the eager and brave to install from the main branch. Installing via pip should look something like

```bash
pipx install --force 'git+https://github.com/conda/conda-lock@main'
```

where `pipx` can be replaced with `pip` and `main` can be replaced with a current Git SHA.

Alternatively, from an `environment.yml`, you can add the following to your dependencies

```yaml
dependencies:
- git  # Only necessary when Git is unavailable at the system level
- pip
- pip:
  - conda-lock @ git+https://github.com/conda/conda-lock@main
```

First draft release notes, up to and including #633:

[explanation of major changes goes here]

[say something nice about pixi, and link to the migration docs]

TODO: categorize the below, remove bot contributions, combine followups, check the authors, e.g. srilman is missing

### Breaking changes

* Update Poetry to 1.8.3 by @romain-intel in https://github.com/conda/conda-lock/pull/637, https://github.com/conda/conda-lock/pull/678, https://github.com/conda/conda-lock/pull/685
* Support Multiple Categories for Sub-Dependencies in Lockfile (Rebase #390) by @srilman in https://github.com/conda/conda-lock/pull/697

### New features

* Sort BaseLockedDependency.dependencies when alpha-sorting a Lockfile by @peterbygrave in https://github.com/conda/conda-lock/pull/654
* Add support for PEP 508 environment markers by @maresb in https://github.com/conda/conda-lock/pull/684
* Improve loading of PyPI mapping by @maresb in https://github.com/conda/conda-lock/pull/690
* Add render-lock-spec subcommand for exporting lock specification to pixi.toml by @maresb in https://github.com/conda/conda-lock/pull/664
* Add pixi.toml configuration by @maresb in https://github.com/conda/conda-lock/pull/705, https://github.com/conda/conda-lock/pull/706
* Add/improve support for Path dependencies and GIT dependencies by @romain-intel in https://github.com/conda/conda-lock/pull/712

### Bugfixes

[...]

PR #753, opened 2024-11-18
Add `--force-platform` to `conda-lock install` command.

jezdez:
### Description

This selects a specific platform for installation instead of the current, native platform. Should be similar to https://github.com/conda/conda/pull/11794, kinda.

### DIFF
--- a/tests/test_conda_lock.py
+++ b/tests/test_conda_lock.py
@@ -2061,16 +2061,16 @@ def test_solve_arch_multiple_categories():
         assert numpy_deps[0].categories == {"test", "dev"}
 
 
-def _check_package_installed(package: str, prefix: str):
-    import glob
-
-    files = list(glob.glob(f"{prefix}/conda-meta/{package}-*.json"))
+def _check_package_installed(package: str, prefix: str, subdir: Optional[str] = None):
+    files = list(glob(f"{prefix}/conda-meta/{package}-*.json"))
     assert len(files) >= 1
     # TODO: validate that all the files are in there
     for fn in files:
         data = json.load(open(fn))
         for expected_file in data["files"]:
             assert (Path(prefix) / Path(expected_file)).exists()
+        if subdir is not None:
+            assert data["subdir"] == subdir
     return True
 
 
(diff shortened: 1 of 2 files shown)


maresb:
Thanks @jezdez! This makes sense.

Some initial thoughts:
1. We should add a test
2. I'm debating whether calling it the mundane `--platform` option is a good idea, since the use case is fairly obscure. Probably it's good for consistency, but I think we should at least indicate in the help string that the option is for forcing an override of the native platform, rather than something routine
3. For internal consistency, how about renaming `install(platform=...)` to `install(force_platform=...)`?

jezdez:
@maresb Yup, this went to the org admins and I just approved it! (more on the PR tomorrow)

jezdez:
> Thanks @jezdez! This makes sense.
> 
> Some initial thoughts:
> 
> 1. We should add a test

I've added one, reusing one of the lockfiles to mimic installing a Intel macOS package on the macOS ARM64 GitHub runners.

> 2. I'm debating whether calling it the mundane `--platform` option is a good idea, since the use case is fairly obscure. Probably it's good for consistency, but I think we should at least indicate in the help string that the option is for forcing an override of the native platform, rather than something routine

I've updated the description

> 3. For internal consistency, how about renaming `install(platform=...)` to `install(force_platform=...)`?

That's fine, I've updated both the argument name as well as the CLI option to make sure it's obvious that it's not quite the same as the `--platform` of the render and lock subcommands.

tests/test_conda_lock.py
```diff
@@ -2650,6 +2650,54 @@ def test_fake_conda_env(conda_exe: str, conda_lock_yaml: Path):
...
+    if Path(conda_lock_yaml).exists():
+        logging.debug(
+            "lockfile contents: \n\n=======\n%s\n\n==========",
+            Path(conda_lock_yaml).read_text(),
+        )
+
+    assert _check_package_installed(
+        package=package,
+        prefix=str(prefix),
+        subdir=platform,
```
jezdez, line 2697: It's checking the subdir value of the metadata here

jezdez:
Thank you @maresb!

leeleavitt:
When will this be released?

maresb:
@leeleavitt, apologies for the delay, you can track the progress towards release in https://github.com/conda/conda-lock/issues/725.

PR #755, opened 2024-11-26
Update miniconda Docker image to use the maintained one.

jezdez:
### Description

I noticed that the docs still use the old, now unmaintained miniconda Docker image, which we shouldn't recommend anymore.

See https://hub.docker.com/r/continuumio/miniconda for more detail.


maresb (approved):
Thanks a lot @jezdez, I really appreciate it!!!

PR #765, opened 2025-01-08
Allow to call `conda-lock -h` (argparse default) instead of `conda-lock --help`

basnijholt:
### Description

After calling `conda-lock -h` MANY times I still fall for it. This adds it as an alias for `--help`.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -1191,8 +1191,10 @@ TLogLevel = Union[
     Literal["CRITICAL"],
 ]
 
+CONTEXT_SETTINGS = {"show_default": True, "help_option_names": ["--help", "-h"]}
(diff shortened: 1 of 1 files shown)


conda_lock/conda_lock.py
```diff
@@ -1191,8 +1191,10 @@ def main() -> None:
     Literal["CRITICAL"],
 ]
 
+CONTEXT_SETTINGS = {"show_default": True, "help_option_names": ["--help", "-h"]}
```
basnijholt, line 1194: I did `["--help", "-h"]` instead of the more common `["-h", "--help"]` to keep the suggestion in the CLI `--help.`

maresb:
Thanks @basnijholt!!!

Issue #641, opened 2024-05-23
incorrect install command suggested in conda-lock.yml comments with 'dev' extra dependencies


qmarcou:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

After building a lock file from a pyproject.toml with optional dependencies named 'dev':
`conda-lock --extras dev -f pyproject.toml --mamba` with ` pyproject.toml` being:
```toml
[project]
name = "MNWE"
dynamic = ["version"]
dependencies = [
    "python",
]

[project.optional-dependencies]
dev = [
    "pytest",
]

[tool.conda-lock]
channels = [
    'conda-forge', 'defaults'
]
platforms = [
    'win-64'
]
```
the conda-lock.yml header displays incorrect install information:
```
# This lock contains optional development dependencies. Include them in the installed environment with:
#     conda-lock install --dev-dependencies -n YOURENV conda-lock.yml
```
which fails
```bash
conda-lock install --dev-dependencies -n YOURENV conda-lock.yml
Usage: conda-lock install [OPTIONS] [LOCK_FILE]
Try 'conda-lock install --help' for help.

Error: No such option: --dev-dependencies
```
Both :
`conda-lock install --dev -n YOURENV conda-lock.yml` and `conda-lock install -e dev -n YOURENV conda-lock.yml` work, any of these two commands should be suggested instead.

I guess these lines should be deleted (if the `-e` option is prefered) or updated:
https://github.com/conda/conda-lock/blob/fcb8d53ee135fd8c473d751343614de1c241a7ab/conda_lock/lockfile/__init__.py#L193-L199

When using any other name than `dev` (e.g `foo`) for extras the correct command is suggested : ``conda-lock install -e foo -n YOURENV conda-lock.yml``

### Conda Info

```shell
active environment : lock-env
    active env location : C:\Users\zzzzz\AppData\Local\miniconda3\envs\lock-env
            shell level : 2
       user config file : C:\Users\zzzzz\.condarc
 populated config files :
          conda version : 24.4.0
    conda-build version : not installed
         python version : 3.10.14.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=skylake_avx512
                          __conda=24.4.0=0
                          __win=0=0
       base environment : C:\Users\zzzzz\AppData\Local\miniconda3  (writable)
      conda av data dir : C:\Users\zzzzz\AppData\Local\miniconda3\etc\conda
[...]

PR #775, opened 2025-02-18
Deprecate --dev related switches

qmarcou:
### Description

This PR is a follow up of #641 .

In a nutshell --dev/--no-dev switches are incorrectly documented, inconsistent, as well as redundant (albeit with some clashes) with the `--extra` CLI.

Changes made deprecate explicitly all --(no-)dev(-dependencies) by raising a meaningful exception.

Limits:
- the deprecation is not in pytest tests. In fact there were no prior tests concerning "--dev" switches, there is a dedicated but unused pytest fixture:
https://github.com/conda/conda-lock/blob/f753f705f38f8166e2dab9365ed5d87d0b90341d/tests/test_conda_lock.py#L356-L364
It seems there is currently no tests regarding the `--extra` switches behavior.
- ongoing discussion about the default behavior of `--extras` in #641

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -1052,8 +1052,7 @@ def _detect_lockfile_kind(path: pathlib.Path) -> TKindAll:
 
 
 _deprecated_dev_help = (
-    "(DEPRECATED) include (or not) dev dependencies in the lockfile (where "
-    "applicable)"
+    "(DEPRECATED) include (or not) dev dependencies in the lockfile (where applicable)"
 )
 
 
@@ -1061,7 +1060,7 @@ def _deprecated_dev_cli(ctx: click.Context, param: click.Parameter, value: Any) 
     """A click callback function raising a deprecation error."""
     if value:
         raise click.BadParameter(
-            "--dev-dependencies/--no-dev-dependencies and --dev/--no-dev "
+            "--dev-dependencies/--no-dev-dependencies (lock, render) and --dev/--no-dev (install) "
             "switches are deprecated. Use `--extra dev` instead."
         )
     else:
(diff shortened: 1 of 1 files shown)


maresb:
Hi @qmarcou, thanks so much for all your work on this. Sorry to take so long to respond, this ended up going a lot deeper than I expected.

I want to move things forward with deprecations, but as I've said I'm also concerned about breaking stuff, especially since the upcoming release is unfortunately going to be so big. The current logic involved is beyond my ability to reason about it, so the only way I could get a handle on it was to test all 48 possible combinations of relevant CLI arguments in https://github.com/conda/conda-lock/pull/778.

I've extended your work a bit, and cherry-picked the tests in https://github.com/qmarcou/conda-lock/pull/2. I would like to know what you think.

maresb:
Thanks so much @qmarcou for coming back to this, I really appreciate it!

I think I've decided on a hard deprecation, but with a flag that can be used as an escape hatch to restore the previous behavior. (I have some uncommitted changes where I think I've correctly replicated the previous behavior.)

Regarding the desired defaults, that's beyond my horizon at the moment. I need to focus fully on releasing conda-lock v3 and not breaking stuff. There's a lot in my head about potential failure modes between mamba v1 and v2, and I still need to make some changes to the test suite.

Once v3 is out (and assuming I'm not inundated with bug reports) then we can immediately figure out the desired default behavior. And now that you know my priorities (simple and clear migration path, and preferring "category" terminology to "extras"), you'll probably be able to figure out the correct solution and let me know once I'm ready. :smile:

maresb:
I want to get v3 out tonight, so I'm going to let this slip. Sorry. :disappointed: 

I plan to release much more frequently in the future, so I don't expect it to be such a big deal to get this in. I sincerely do value your contribution and feedback.

qmarcou:
No worries, and congratulations for getting v3 out!
Apologies I have been quite busy lately, and I wasn't sure whether you were waiting for action on my part before merging https://github.com/qmarcou/conda-lock/pull/2 in this current PR.
I have just unlocked the workflow runs, and I could chnage the UserWarning to FutureWarning and we should be good to go?

maresb:
Thanks @qmarcou!

Supposing we fix the warning and merge my PR, there are still two things on my mind:
1. Should we also merge #778?
2. I think my PR was triggering some spurious deprecation warnings when running without any arguments, so I think my detection for the distinction between omitted flags and explicitly-specified default flags may still be broken.

PR #776, opened 2025-02-18
Use build_string in build field in fake package

RSchwan:
### Description

Concatenating the`build` field in a conda package as `{build_string}_{build_number}` is wrong and leads to wrong behaviors.

For example, running `conda-lock -f environment.yml` with
```yaml
# environment.yml
channels:
  - conda-forge
dependencies:
  - libblasfeo>=0.1.4.1
platforms:
  - linux-64
```
result in the following error:
```
{
    "solver_problems": [
        "nothing provides __archspec 1 core2 needed by _x86_64-microarch-level-1-0_core2"
    ],
    "success": false
}
```
The error occurs because `FakePackage(name="__archspec", version="1", build_string="x86_64")` ([see here](https://github.com/conda/conda-lock/blob/df4280c4562239b6a0aa23517416f0402726d80a/conda_lock/virtual_package.py#L195)) has the `build` field set to `x86_64_0` which leads to the error since it would expect `__archspec-1-x86_64.tar.bz2` (or `__archspec-1-core2.tar.bz2` as an alternative example) as a package instead of `__archspec-1-x86_64_0.tar.bz2`. This becomes especially critical, since the `_x86_64-microarch-level-*` packages need exact build fields for `__archspec`.

#### Edit 

Also added some test and added an example for `__archspec` in the docs.

### DIFF
--- a/conda_lock/virtual_package.py
+++ b/conda_lock/virtual_package.py
@@ -37,9 +37,9 @@ class FakePackage(BaseModel):
     def to_repodata_entry(self) -> Tuple[str, Dict[str, Any]]:
         out = self.model_dump()
         if self.build_string:
-            build = f"{self.build_string}_{self.build_number}"
+            build = self.build_string
         else:
-            build = f"{self.build_number}"
+            build = str(self.build_number)
         out["depends"] = list(out["depends"])
         out["build"] = build
         fname = f"{self.name}-{self.version}-{build}.tar.bz2"
--- a/README.md
+++ b/README.md
@@ -257,9 +257,11 @@ subdirs:
     packages:
       __glibc: "2.17"
       __cuda: "11.4"
+      __archspec: "1 haswell"
   win-64:
     packages:
       __cuda: "11.4"
+      __archspec: "1 haswell"
 ```
 
 conda-lock will automatically use a `virtual-packages.yml` it finds in the the current working directory.  Alternatively one can be specified
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -147,9 +147,11 @@ subdirs:
     packages:
       __glibc: "2.17"
       __cuda: "11.4"
+      __archspec: "1 haswell"
   win-64:
     packages:
       __cuda: "11.4"
+      __archspec: "1 haswell"
 ```
 
 conda-lock will automatically use a `virtual-packages.yml` it finds in the the current working directory.  Alternatively one can be specified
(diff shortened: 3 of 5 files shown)


conda_lock/virtual_package.py
```diff
@@ -37,9 +37,9 @@ class FakePackage(BaseModel):
     def to_repodata_entry(self) -> Tuple[str, Dict[str, Any]]:
         out = self.model_dump()
         if self.build_string:
-            build = f"{self.build_string}_{self.build_number}"
+            build = self.build_string
```
maresb, line 40: Thanks so much for this!!!!!

I'm struggling to wrap my mind around this part. In particular, if the original was wrong, then why didn't it break more stuff prior to this?

You write,

> This becomes especially critical, since the `_x86_64-microarch-level-*` packages need exact build fields for `__archspec`.

Does this mean that other packages don't need exact build fields? Is there some sort of wildcard matching that makes it irrelevant for other cases?

Thanks so much for your patience.

maresb:
Thanks also for the test cases! :star_struck:

conda_lock/virtual_package.py
RSchwan: I'm not 100% sure, but as I understand it from the [documentation](https://docs.conda.io/projects/conda-build/en/stable/resources/package-spec.html#package-match-specifications), the build flag has to match exactly. But most packages don't depend on a build flag, though. For example, `libblasfeo` has the following dependencies
```json
"depends": [
    "__glibc >=2.17,<3.0.a0",
    "_x86_64-microarch-level >=1",
    "libgcc >=13"
  ],
```
and `_x86_64-microarch-level-3-2_x86_64_v3` has the following dependencies
```
"depends": [
    "__archspec 1 x86_64_v3"
  ],
```
which requires an exact build flag of `x86_64_v3`. This means that for most packages, the build field is completely ignored, which is why it probably was never an issue. But in the special cases where it has to match, like in the case of `__archspec`, this was a problem.

RSchwan: Another note to clarify this. If you for example take a look at the numpy packages [here](https://anaconda.org/conda-forge/numpy/files), you can see that the build flag is actually generated for most packages based on a random string and the build_number. But again, since most packages only depend on an explicit version and not a build flag, this is not an issue.

maresb (approved):
Thanks for this, I think I understand now, and it looks good. Is this ready to be merged? (If I don't hear back soon, then I'm going to assume yes.)

RSchwan:
Can be merged from my side.

PR #789, opened 2025-03-29
Generate lockfiles for separate Python versions and test accordingly

### LINKED ISSUE #783: CI runs intended for Python 3.9 actually use Python 3.12 instead
Oops, this one's pretty embarrassing.

We'll need separate lockfiles for each Python version we test.

The strategy I have in mind is to create separate environment-python-3.X.yml files for each Python version, and then run the lock command on both (simultaneously) the primary environment.yml and environment-python-3.X.yml.

That description was probably confusing, so it's probably easier to look at the implementation I did in https://github.com/pymc-labs/CausalPy/pull/281.

With this scheme I find it slightly awkward to maintain separate files just to specify the desired Python version, but ultimately I think it should be very little effort. (I'm open to alternative suggestions.)

**RELATED**: If we want the upcoming release to support micromamba v1, we should add that to the matrix as well.

maresb:
### Description

Closes #783


maresb:
Hmm... Mamba v1 isn't available yet for Python 3.13.

https://github.com/conda-forge/mamba-feedstock/issues/309

PR #792, opened 2025-04-04
Ensure `--check-input-hash` is respected when using `conda-lock lock --strip-auth --check-input-hash ...`

peterbygrave:
### Description
When you use `conda-lock lock --strip-auth --check-input-hash` it always tries to re-lock the file. I found this because when trying to diagnose difference between local solve and CI - but that was a different matter.

I have attempted to fix this but I don't know how to test, or if there is a reason for it.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -1430,6 +1430,7 @@ def lock(
         channel_overrides=channel_overrides,
         kinds=kind,
         lockfile_path=None if lockfile is None else pathlib.Path(lockfile),
+        check_input_hash=check_input_hash,
         extras=extras_,
         virtual_package_spec=virtual_package_spec,
         with_cuda=with_cuda,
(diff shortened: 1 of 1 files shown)


maresb (approved):
LGTM, thanks a lot @peterbygrave!

@mariusvniekerk, did you have any particular motivation for omitting `check_input_hash` when `strip-auth=False` in https://github.com/conda/conda-lock/pull/104?

Regarding tests, we certainly can't test all possible combinations of parameters, so unless there's a better idea, I'd just skip tests on this one.

Now that CI is finally green, I'm eager to cut a patch release for this. :smile:

maresb:
While waiting for a response from @mariusvniekerk, I've had a few thoughts:

1. I think the underlying code could be refactored in order to make this type of bug less likely. In particular, I'd prefer that the Click `lock` function sticks to parsing the CLI args and leaves the logic for things like auth stripping to the `run_lock` function.
2. It would be really nice to have more thorough integration tests so that we actually exercise things like `--check-input-hash` with authenticated repositories.

Both of these seem a bit out-of-scope for this PR though.

I'll wait another day or two for feedback from Marius before merging this.

Issue #432, opened 2023-06-12
Remove the `content_hash`s?


baszalmstra:
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

While working on porting the conda lock format to Rust I'm having a tough time implementing the `content_hash`. The implementation relies heavily on the serialization of the model to json which is hard to get precisely compatible with `conda-lock`.

In the broader sense, I wonder what the use of `content_hash` is. Is it to verify the integrity of the file?

### Why is this needed?

To make integrating the conda-lock file format in other languages much easier.

### What should happen?

Basically, drop the `content_hash` field from the format.

maresb:
Ya, I think it's mainly to ensure that the file hasn't been hand-modified. Personally, I vaguely think it's good, but I don't have strong feelings. I think @mariusvniekerk feels a bit more strongly.

See https://github.com/mamba-org/mamba/issues/1209#issuecomment-944521460 for more context

baszalmstra:
I think that discussion is more about the input specification than about the content of the lock file itself. The `content-hash` is currently a hash of the `LockedDependency`s if I understand it correctly.

Additionally, I also don't think including a hash of the input provides much value. First of all you can easily check if the contents of the lock-file is "valid" for a given input. Simply check for each input matchspec if there is a locked package in the lockfile that matches it. You can even validate that the entire lock-file is consistent by also traversing the dependencies of the locked packages.

Secondly, only hashing the input specification is not enough to create a reproducible conda-lock file since the `repodata` or the micromamba version and stuff like that are not included.

baszalmstra:
@mariusvniekerk Are you talking about the `input_hash` or the `content_hash`. If I read it correctly the `content_hash` is not based on the input but based on the actual locked-down packages, thus simply computing the sha256 of the entire file would likely yield similar results. 

> The input_hash comes somewhat from a time before mamba where recomputing a lock was very expensive and this was a cheap way to know whether to even bother attempting to recompute it.

You could also very cheaply do the same thing by checking if the input already matches the locked-down packages. This can all be done in Python without ever touching a solver.

My concern is that computing hashes is a fragile thing. If you add/remove a field, or format it slightly differently the hashes will not match. This makes it very hard to reproduce in a different environment or in another language or iterate on the design. I think doing semantic checks over checking hashes is a more stable approach. 

[...]

PR #797, opened 2025-05-02
[WIP] Improve compatibility between lockfiles produced by v2 and v3

maresb:
### Description

The `content_hash` changes upon upgrading from v2 to v3 when it shouldn't. We should fix this.


maresb:
Obligatory reminder that `content_hash` is fundamentally flawed: https://github.com/conda/conda-lock/issues/432#issuecomment-1637071282.

Unfortunately I don't see an easy/obvious way to check satisfiability here in conda-lock. We really should switch at some point.

maresb:
I'm gradually making the `content_hash` computation more functional (in the sense of functional programming) so that we can reproduce and test both the v2 hashes and the regressed v3.0.0 and v3.0.1 hashes.

maresb:
I think this solves it. Will take another pass to double-check before releasing.

Issue #786, opened 2025-03-28
Minor: flaky test_concurrent_cached_download_file


maresb:
On a few occasions I have noticed `test_concurrent_cached_download_file` failing with the following error.

I want to start collecting occurrences of this failure here so that I can, for example, see if the issue is operating-system specific.

```python
>           assert (
                thread_names_calling_requests_get.qsize()
                == 1
                == len(set(thread_names_calling_requests_get.queue))
                == mock_get.call_count
            ), f"{thread_names_calling_requests_get.queue=}"
E           AssertionError: thread_names_calling_requests_get.queue=deque(['CachedDownloadFileThread-1', 'CachedDownloadFileThread-0'])
E           assert 2 == 1
E            +  where 2 = qsize()
E            +    where qsize = <queue.Queue object at 0x10babbef0>.qsize

tests/test_lookup_cache.py:308: AssertionError
```

What's happening with this test failure is that I have a `FileLock` intended to ensure that only one concurrent process (for testing I'm using threads instead of processes) downloads the lookup table. Somehow the lock is failing and two threads download the lookup table.

This is not particularly serious, because it works most of the time, and when it fails it only leads to an extra unnecessary download.

What's especially tricky is that this is an occasional flake so I don't have a reliable reproducer.

## Failed attempt

I was thinking that maybe I need to [set `thread_local=False`](https://py-filelock.readthedocs.io/en/latest/index.html#filelocks-and-threads) to make the `FileLock` global across threads. However, I now doubt that this is the correct fix:

As far as I can tell, what `thread_local=False` does is to make it so that, if the same `FileLock` instance (as a Python object) is shared between multiple threads, then each thread can simultaneously acquire the lock. If each thread creates its own `FileLock` instance then I don't think `thread_local` has any effect.

Based on my belief that `thread_local` has no effect, I don't think it's justified to introduce it.

Here's a script to demonstrate the `thread_local` behavior:

```python
import threading
import time
from filelock import FileLock

def worker(name: str, lock: FileLock):
    print(f"Thread {name} attempting to acquire the lock.")
    with lock:
        print(f"Thread {name} acquired the lock!")
        time.sleep(2)
        print(f"Thread {name} releasing the lock.")
    print(f"Thread {name} finished.")

if __name__ == "__main__":
    for thread_local in [True, False]:
        for same_lock_object in [True, False]:
            lock_a = FileLock("test.lock", thread_local=thread_local)
            if same_lock_object:
                lock_b = lock_a
[...]

PR #811, opened 2025-06-21
Refactor lookup cache test to use multiprocessing instead of threading

maresb:
I occasionally observe some odd failures of the lookup cache test, already detailed in https://github.com/conda/conda-lock/issues/786. As far as I can tell, everything is working correctly except for the test. I wonder if the test would be more reliable using multiprocessing in place of multithreading. After all, multiprocessing is the type of concurrency we actually use. (Threads are just simpler than processes, and I thought they would be sufficient.) Let's see if this improves the situation.

### DIFF
--- a/conda_lock/conda_lock.py
+++ b/conda_lock/conda_lock.py
@@ -1160,7 +1160,6 @@ def run_lock(
     _conda_exe = determine_conda_executable(
         conda_exe, mamba=mamba, micromamba=micromamba
     )
-    logger.debug(f"Using conda executable: {_conda_exe}")
     version_info = subprocess.check_output(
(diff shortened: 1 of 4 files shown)


maresb:
It didn't work, I just got the following [on Windows](https://github.com/conda/conda-lock/actions/runs/15834962407/job/44636306210?pr=805):

```
 E           AssertionError: process_names_emitting_lock_warnings_list=['CachedDownloadFileProcess-0', 'CachedDownloadFileProcess-1', 'CachedDownloadFileProcess-3']
E           assert 3 == 4
E            +  where 3 = len(['CachedDownloadFileProcess-0', 'CachedDownloadFileProcess-1', 'CachedDownloadFileProcess-3'])
```

PR #815, opened 2025-06-27
Don't clobber build extensions (.tar.bz2 vs .conda) when running conda-lock update

### LINKED ISSUE #639: bz2 suffixes being used despite registry not having bz2 files
### Checklist

### What happened?

Using the `--update` flag seems to result in all of my packages that have `.conda` suffixes being replaced with `.tar.bz2` suffixes. It seems to because of [this line](https://github.com/conda/conda-lock/blob/53eeb696f29121890d63bf8a5d39c87e5580d762/conda_lock/conda_solver.py#L495).

I took a look at fixing it but it seems like the overall strategy being used here of running `conda list --json` won't work since it doesn't give us the remote URL and so we're forced to re-assemble it from the channel and guess at what the extension is.

maresb:
### Description

Fixes #639


maresb:
I would still like to see if this has any effect on pip dependencies

PR #847, opened 2025-10-08
Add 2.18 as a valid glibc version

romain-intel:
### Description

Some packages like `nvidia-nccl-cu12` list 2.18 as their minimum GLIBC version. Conda-lock would only add 2.17 (which is more standard I guess) causing resolution to fail.

An example failing environment is the following:

```
channels:
  - conda-forge
dependencies:
  - numpy <=2
  - pip:
    - xgboost >=2
```

`xgboost` indirectly uses `nvidia-nccl-cu-12` which fails. This tiny patch just adds 2.18 as a GLIBC option if supported and makes the above environment properly resolvable.

### DIFF
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -60,7 +60,7 @@ if TYPE_CHECKING:
 # in practice, see https://github.com/pypa/manylinux/blob/main/README.rst#docker-images.
 # NOTE:
 #   Keep the max in sync with the default value used in default-virtual-packages.yaml.
-MANYLINUX_TAGS = ["1", "2010", "2014", "_2_17", "_2_24", "_2_28"]
+MANYLINUX_TAGS = ["1", "2010", "2014", "_2_17", "_2_18", "_2_24", "_2_28"]
 
 # This needs to be updated periodically as new macOS versions are released.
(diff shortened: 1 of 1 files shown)


maresb:
The conda-forge default seems to be 2.17, set [here](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/de3a0e060b8fa84dc7eb2fa492050daed742364b/recipe/conda_build_config.yaml#L19).

It seems that we support down to 2.5, but only via the legacy aliases.

Our previous idea was to support the [tags](https://quay.io/organization/pypa) generated by the [PyPA manylinux project](https://github.com/pypa/manylinux/blob/main/README.rst#docker-images). Curiously, 2.18 doesn't seem to be in that list.

So maybe we should be doing:

```python
["_2_5", "1", "_2_6", "_2_7", "_2_8", "_2_9", "_2_10", "_2_11", "_2_12", "2010", "_2_13", "_2_14", "_2_15", "_2_16", "_2_17", "2014", "_2_18", "_2_19", "_2_20", "_2_21", "_2_22", "_2_23", "_2_24", "_2_25", "_2_26", "_2_27", "_2_28"]
```

maresb:
Merging for now, but still would like to know if we can do better.

PR #900, opened 2026-03-17
Add manylinux_2_34 to supported MANYLINUX_TAGS

guenp:
## Summary

- Adds `_2_34` to `MANYLINUX_TAGS` in `pypi_solver.py`
- Bumps default `__glibc` from `2.28` to `2.34` in `default-virtual-packages.yaml` for all Linux subdirs

Follow-up to #847, which added `_2_18`.

## Motivation

[pyqir 0.12.3](https://pypi.org/project/pyqir/0.12.3/#files) only publishes Linux wheels for `manylinux_2_34` (x86_64) and `manylinux_2_38` (aarch64). Since `MANYLINUX_TAGS` currently maxes out at `_2_28`, conda-lock's PyPI solver rejects these wheels:

```
RuntimeError: Unable to find installation candidates for pyqir (0.12.3)
```

The `--virtual-package-spec` workaround doesn't help because even if `__glibc` is set to `2.34`, the hardcoded tag list doesn't include `_2_34`.

## Discussion

I'm open to discussing a better long-term solution — e.g., dynamically generating valid manylinux tags from the glibc version rather than maintaining a hardcoded list. Happy to iterate on this.

## Test plan

- [ ] Verify pyqir 0.12.3 resolves successfully for linux-64
- [ ] Verify `--virtual-package-spec` with `__glibc: "2.34"` works end-to-end
- [ ] Verify backward compatibility (packages with older manylinux tags still resolve)

### DIFF
--- a/conda_lock/content_hash.py
+++ b/conda_lock/content_hash.py
@@ -159,6 +159,43 @@ def _virtual_package_content_for_platform(
     return result
 
 
+def _replace_glibc_version(
+    virtual_package_repo: FakeRepoData,
+    platform: PlatformSubdirStr,
+    target_version: str,
+) -> FakeRepoData | None:
+    """Create a VPR variant with a different glibc version.
+
+    Returns None if the platform has no glibc package or already matches.
+    """
+    import copy
+
+    repodata = virtual_package_repo.all_repodata
+    if platform not in repodata:
+        return None
+    subdir_data = repodata[platform]
+    if not isinstance(subdir_data, dict) or "packages" not in subdir_data:
+        return None
+    packages = subdir_data["packages"]
+    # Find the glibc package key
+    glibc_keys = [k for k in packages if k.startswith("__glibc-")]
+    if not glibc_keys:
+        return None
+    glibc_key = glibc_keys[0]
+    current_version = packages[glibc_key].get("version")
+    if current_version == target_version:
+        return None
+
+    new_repo = copy.deepcopy(virtual_package_repo)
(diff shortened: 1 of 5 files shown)


guenp:
@maresb curious to hear your thoughts on this PR. thank you!

maresb:
Hi @guenp, apologies for the late review.

Looking this over, I think this PR became overcomplicated due to a preexisting flaw in our tests going back to #566. I document and attempt to fix this flaw in #903. Would you please look that over and see if that makes sense to you?

Regarding this PR, assuming that we merge #903, then I think it's sufficient as a temporary fix to simply add the corresponding 2.34 entry to MANYLINUX_TAGS. I think this would be a 1-line PR. (Sorry for all the unnecessary complexity you faced in the current draft, but at least this led me to open #903.)

As for a more sustainable solution, I think we can just remove `MANYLINUX_TAGS` and instead simplify `_compute_compatible_manylinux_tags` to just generate the list up to whatever arbitrary value is specified in the virtual packages. (I think the only detail is injecting the special aliases in `SPECIAL_CASES`.) So unless I'm missing something, this should be a pretty big simplification of the current logic. (I think this implementation would be the logical conclusion to the discussions in #847.)

Is this something you'd be interested in tackling?

maresb:
@romain-intel, if you're available I'd love to get your take on this!

guenp:
Thanks, @maresb for reviewing and fixing the test issue! That sounds good to me, I'm happy to contribute the `_compute_compatible_manylinux_tags` solution after #903 is merged.

PR #905, opened 2026-04-08
fix: register as conda plugin

### LINKED ISSUE #904: command line integration breaks w/ conda 26.3.0
The latest conda removed the feature to pickup subcommanda via `conda-*` commands. This is breaking conda-lock which needs to now use the proper plugin.

cc @kenodegard

beckermr:
### Description

The latest conda versions switched to plugin hooks for custom subcommands (e.g., `conda lock`) and as of conda 26.3, the old interface was deprecated. This PR adds the proper plugin hooks.

To Do:

- [x] add plugin hooks
    - [x] add basic code for the hook
    - [x] figure out how to pass argv to click per the todo item
- [x] adjust logging config so we don't rely on configuring the root logger
- [x] add tests w/ conda in the python env to ensure the new plugin hooks work as expected

closes #904

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -39,6 +39,10 @@ jobs:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
         use_conda_cli: ["false", "true"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.14"
+            use_conda_cli: "true"
     steps:
       - name: Set Conda platform
         run: |
@@ -119,6 +123,10 @@ jobs:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
         use_conda_cli: ["false", "true"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.14"
+            use_conda_cli: "true"
     steps:
       - name: Set Conda platform
         run: |
@@ -169,6 +177,10 @@ jobs:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
         use_conda_cli: ["false", "true"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.14"
+            use_conda_cli: "true"
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -217,6 +229,10 @@ jobs:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
         use_conda_cli: ["false", "true"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.14"
+            use_conda_cli: "true"
     steps:
       - uses: conda-incubator/setup-miniconda@fc2d68f6413eb2d87b895e92f8584b5b94a10167 # v3.3.0
         with:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -86,7 +86,7 @@ jobs:
         if: matrix.use_conda_cli == 'true'
         shell: bash -leo pipefail {0}
         run: |
-          conda install pip setuptools hatchling hatch-vcs conda-lock --yes
+          conda install pip setuptools editables hatchling hatch-vcs conda-lock --yes
           conda uninstall conda-lock --force --yes
           pip install --no-deps --no-build-isolation -e .
       - name: run-test
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -38,11 +38,6 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
-        use_conda_cli: ["false", "true"]
-        exclude:
-          - os: windows-latest
-            python-version: "3.14"
-            use_conda_cli: "true"
     steps:
       - name: Set Conda platform
         run: |
@@ -60,17 +55,14 @@ jobs:
         with:
           persist-credentials: false
       - name: Set up Python
-        if: matrix.use_conda_cli == 'false'
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: ${{ matrix.python-version }}
       - name: install conda-lock
-        if: matrix.use_conda_cli == 'false'
         run: |
           pip install pipx
           pipx install -e .
       - name: run-test
-        if: matrix.use_conda_cli == 'false'
         run: |
           conda-lock --log-level=DEBUG --micromamba -f tests/gdal/environment.yml -p $CONDA_PLATFORM
           conda-lock render -p $CONDA_PLATFORM
@@ -80,35 +72,10 @@ jobs:
           cat conda-lock.yml
           mkdir lockfiles
           mv conda-$CONDA_PLATFORM.lock conda-$CONDA_PLATFORM.lock.yml conda-lock.yml lockfiles
-      - name: Set up Python
-        if: matrix.use_conda_cli == 'true'
-        uses: conda-incubator/setup-miniconda@fc2d68f6413eb2d87b895e92f8584b5b94a10167 # v3.3.0
-        with:
-          auto-activate-base: true
-          python-version: ${{ matrix.python-version }}
-      - name: install conda-lock
-        if: matrix.use_conda_cli == 'true'
-        shell: bash -leo pipefail {0}
-        run: |
-          conda install conda pip setuptools editables hatchling hatch-vcs conda-lock --yes
-          conda uninstall conda-lock --force --yes
-          pip install --no-deps --no-build-isolation -e .
-      - name: run-test
-        if: matrix.use_conda_cli == 'true'
-        shell: bash -leo pipefail {0}
-        run: |
-          python -m conda lock --log-level=DEBUG --micromamba -f tests/gdal/environment.yml -p $CONDA_PLATFORM
-          python -m conda lock render -p $CONDA_PLATFORM
-          python -m conda lock render -p $CONDA_PLATFORM --kind=env
-          cat conda-$CONDA_PLATFORM.lock
-          cat conda-$CONDA_PLATFORM.lock.yml
-          cat conda-lock.yml
-          mkdir lockfiles
-          mv conda-$CONDA_PLATFORM.lock conda-$CONDA_PLATFORM.lock.yml conda-lock.yml lockfiles
       - name: Upload lockfiles
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
-          name: conda-lock-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.use_conda_cli }}
+          name: conda-lock-${{ matrix.os }}-${{ matrix.python-version }}
           path: lockfiles
 
   install-gdal-with-micromamba:
@@ -122,11 +89,6 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ "3.10", "3.14" ]
-        use_conda_cli: ["false", "true"]
-        exclude:
-          - os: windows-latest
-            python-version: "3.14"
-            use_conda_cli: "true"
     steps:
       - name: Set Conda platform
         run: |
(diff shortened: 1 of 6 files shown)


beckermr:
pre-commit.ci autofix

beckermr:
Ah maybe this was never meant to be a subcommand since `conda` is not listed in the dependencies in the pyproject.toml. 

@maresb @mariusvniekerk what is the history on `conda lock` vs. `conda-lock`?

maresb:
Hey @beckermr! `lock` is the default subcommand of `conda-lock`. I'm not so familiar how `conda` handles `lock` as a subcommand, but it seems reasonable that it should work when installed so that `conda-lock` is equivalent to `conda lock`. It would probably make sense to allow `conda lock conda-lock-subcommand` so that for example `conda lock install` would work. Is this what you mean? I haven't yet looked into your linked issue.

Also, I'm not sure what's up with the CI, but I noticed similar failures last night, so I think it's unrelated to your PR. I'll try to investigate soon.

beckermr:
OK. If we merge this PR, then we have to permanently add conda as a dependency of conda-lock. Is that OK?

maresb:
> OK. If we merge this PR, then we have to permanently add conda as a dependency of conda-lock. Is that OK?

No, that would break the current pip-installability of conda-lock.

Does the entry-point actually require `conda` to be a dependency?

beckermr:
It currently needs this import `from conda.plugins import hookimpl`. Maybe I can catch the error and not register? Let me try it.

beckermr:
I'll need to add tests here.

conda_lock/plugin.py
```diff
@@ -0,0 +1,35 @@
...
+
+        Lazy import to avoid import-time side effects when not using conda-lock.
+        """
+        from conda_lock.__main__ import main
+
+        return main()  # TODO: does not accept args for parsing since using click
+
+
+    @hookimpl
+    def conda_subcommands():
```
beckermr:
```suggestion
    def conda_subcommands() -> CondaSubcommand:
```

beckermr:
pre-commit.ci autofix

beckermr:
Any hints on the `ty` test failure @maresb? In principle it requires conda to be installed but that would mean reconfiguring the CI job to not use pypi for the typechecks. I'd rather not do that...

maresb:
Let's use `# type: ignore` here.

beckermr:
LOL that broke mypy 🤦

conda_lock/plugin.py
```diff
@@ -0,0 +1,36 @@
...
+Registers ``conda lock`` so the feedstock tooling is available as a conda
+subcommand when this package is installed in the same environment as conda.
+"""
+
+from __future__ import annotations
+
+
+try:
+    from conda.plugins import hookimpl  # type: ignore
+    from conda.plugins.types import CondaSubcommand  # type: ignore
```
beckermr:
```suggestion
    from conda.plugins import hookimpl  # type: ignore[unused-ignore]
    from conda.plugins.types import CondaSubcommand  # type: ignore[unused-ignore]
```

maresb:
LGTM, I just need to fix the broken CI. Thanks @beckermr!

beckermr:
xref: https://github.com/regro/cf-scripts/pull/5889 for the python-build failure

beckermr:
This needs integration tests to ensure the `conda lock` entry point works as expected. I will add those.

beckermr:
We need to remove all of the logging.basicConfig calls in order for the plugin to work properly with conda. see https://github.com/conda/conda/issues/15872

maresb:
> We need to remove all of the logging.basicConfig calls in order for the plugin to work properly with conda. see [conda/conda#15872](https://github.com/conda/conda/issues/15872)

Oof, that sounds like a pain. Probably something for a separate PR. Maybe I can get an agent to sort it out.

beckermr:
Not too bad. I just did this for smithy. I will push it here.

beckermr:
xref: https://github.com/conda-forge/conda-smithy/pull/2512

beckermr:
pre-commit.ci autofix

conda_lock/common.py
```diff
@@ -66,3 +67,13 @@ def relative_path(source: pathlib.Path, target: pathlib.Path) -> str:
 
 def warn(msg: str) -> None:
     warnings.warn(msg, stacklevel=2)
+
+
+def configure_logger_basic(logger: logging.Logger, level: int = logging.INFO): -> None:
```
beckermr:
```suggestion
def configure_logger_basic(logger: logging.Logger, level: int = logging.INFO) -> None:
```

conda_lock/common.py
```diff
@@ -66,3 +67,13 @@ def relative_path(source: pathlib.Path, target: pathlib.Path) -> str:
 
 def warn(msg: str) -> None:
     warnings.warn(msg, stacklevel=2)
+
+
+def configure_logger_basic(logger: logging.Logger, level: int = logging.INFO) -> None:
```
beckermr:
```suggestion
def configure_logger_basic(logger: logging.Logger, level: int | Literal['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'] = logging.INFO) -> None:
```

conda_lock/common.py
beckermr, line 9:
```suggestion
    Any,
    Literal,
```

beckermr:
pre-commit.ci autofix

maresb (approved):

maresb:
Looks great @beckermr!

Are you still working on this?

beckermr:
Yes! Let me add a todo list at the top.

conda_lock/plugin.py
```diff
@@ -0,0 +1,36 @@
...
+if HAVE_CONDA:
+
+    def _execute(args: tuple[str, ...]) -> int | None:
+        """Dispatch plugin arguments to the lock CLI.
+
+        Lazy import to avoid import-time side effects when not using conda-lock.
+        """
+        from conda_lock.__main__ import main
+
+        return main()  # TODO: does not accept args for parsing since using click
```
beckermr: I need to figure out how to pass `argv` to a click CLI.

maresb: Unvalidated generated suggestion:

The fix is straightforward. Click commands accept `args` as the first positional parameter, and `standalone_mode=False` prevents click from calling `sys.exit()` (returning the exit code instead), which is what conda expects from the action callback (typed as `Callable[[tuple[str, ...]], int | None]`):

```python
def _execute(args: tuple[str, ...]) -> int | None:
    from conda_lock.conda_lock import main
 
    return main(list(args), standalone_mode=False)
```

Key details:

*    list(args) — click expects a `list` (or `None`), not a tuple. When provided, click uses these instead of `sys.argv[1:]`.
*    `standalone_mode=False` — prevents click from calling `sys.exit()` after execution. Instead, it returns the command's return value on success, and returns the exit code if a `SystemExit` is raised (e.g. from `--help` or `--version`). Click exceptions like `UsageError` propagate as normal exceptions.
*    Import from `conda_lock.conda_lock` rather than `conda_lock.__main__`, since `__main__.py` calls `main()` with no args (triggering `sys.argv` parsing and `sys.exit`).

This matches the pattern from conda-smithy, where `main(args)` passes `args` through to `argparse.parse_args(args)` — it's just the click equivalent.

beckermr: Oh cool. I will try it out locally.

beckermr: This works!

```
% python -m conda lock
ERROR:conda_lock.conda_lock:No source files provided and no default files found. Exiting.
No source files provided and no default files found. Exiting.

% python -m conda lock --help
Usage: python -m conda [OPTIONS] COMMAND [ARGS]...

  To get help for subcommands, use the conda-lock <SUBCOMMAND> --help

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  lock*             Generate fully reproducible lock files for conda...
  install
  render            Render multi-platform lockfile into single-platform...
  render-lock-spec  Combine source files into a single lock specification
% python -m conda lock lock --help
Usage: python -m conda lock [OPTIONS]

  Generate fully reproducible lock files for conda environments.

  By default, a multi-platform lock file is written to conda-lock.yml.

  When choosing the "explicit" or "env" kind, lock files are written to
  conda-{platform}.lock. These filenames can be customized using the
  --filename-template argument. The following tokens are available:

      platform: The platform this lock file was generated for (conda subdir).
      dev-dependencies: Whether or not dev dependencies are included in this lock file.
      input-hash: A sha256 hash of the lock file input specification.
      version: The version of conda-lock used to generate this lock file.
      timestamp: The approximate timestamp of the output file in ISO8601 basic format.

Options:
  --conda TEXT                    path (or name) of the conda/mamba executable
                                  to use.
  --mamba / --no-mamba            don't attempt to use or install mamba.
                                  [default: mamba]
  --micromamba / --no-micromamba  don't attempt to use or install micromamba.
                                  [default: no-micromamba]
  -p, --platform TEXT             generate lock files for the following
                                  platforms
```

conda_lock/plugin.py
```diff
@@ -0,0 +1,36 @@
...
+if HAVE_CONDA:
+
+    def _execute(args: tuple[str, ...]) -> int | None:
+        """Dispatch plugin arguments to the lock CLI.
+
+        Lazy import to avoid import-time side effects when not using conda-lock.
+        """
+        from conda_lock.conda_lock import main
+
+        return main(args, standalone_mode=False)
```
beckermr, line 29:
```suggestion
        # code below suggested by AI and appears to work
        from conda_lock.conda_lock import main

        return main(args, standalone_mode=False)
```

beckermr:
OK @maresb! This one is ready for review!

beckermr:
This approach is causing duplicate log messages in smithy. I am going to try a fix there and then update this PR if that works.

beckermr:
ok @maresb - This is ready for review again. The PR is a lot simpler now.

.github/workflows/integration-test.yml
```diff
@@ -182,7 +227,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ "3.10" ]
+        python-version: [ "3.10", "3.14" ]
+        use_conda_cli: ["false", "true"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.14"
+            use_conda_cli: "true"
```
maresb: Why this exclude? Does setup-miniconda not work yet with python 3.14? Could we add a comment explaining why?

.github/workflows/integration-test.yml
```diff
@@ -72,11 +80,37 @@ jobs:
           cat conda-lock.yml
           mkdir lockfiles
           mv conda-$CONDA_PLATFORM.lock conda-$CONDA_PLATFORM.lock.yml conda-lock.yml lockfiles
+      - name: Set up Python
```
maresb:
```suggestion
      - name: Set up Conda
```

maresb (commented):
Sorry it took me a while to review this.

Beautiful simplification with the logging!

My main concern is how this increases the total number of integration test jobs from 18 to 44. I have the impression that some type of API limits cause runner allocation to stall, especially for the macos runners. Thus I'm worried that this increase could lead to major bottlenecks for how long it takes the conda-lock CI to complete.

Would it be sufficient to add a separate job that installs Conda and runs something trivial like `conda lock --help` to revert the matrix expansion?

beckermr:
OK @maresb I reduced the integration test to one extra job. This is ready for another look!

.github/workflows/integration-test.yml
```diff
@@ -87,7 +88,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ "3.10" ]
+        python-version: [ "3.10", "3.14" ]
```
beckermr:
```suggestion
        python-version: [ "3.10" ]
```

.github/workflows/integration-test.yml
```diff
@@ -182,7 +184,7 @@ jobs:
...
```
beckermr:
```suggestion
        python-version: [ "3.10" ]
```

.github/workflows/integration-test.yml
```diff
@@ -77,6 +77,7 @@ jobs:
         with:
           name: conda-lock-${{ matrix.os }}-${{ matrix.python-version }}
           path: lockfiles
+
```
beckermr:
```suggestion
```

.github/workflows/integration-test.yml
```diff
@@ -125,6 +126,7 @@ jobs:
           micromamba run -n gdal-test-mm-lock python -c 'import toolz; print(toolz.__version__)'
           # Micromamba won't install pip dependencies from explicit lockfiles, so skip this test
           micromamba run -n gdal-test-mm-env python -c 'import toolz; print(toolz.__version__)'
+
```
beckermr:
```suggestion
```

maresb (approved):
Looks great @beckermr!

PR #920, opened 2026-06-26
Merge duplicate dependency specs

basnijholt:
Some solver results contain duplicate dependency specs for the same package, e.g. both `python` and `python >=3.10` for a noarch package. The current dict assignment keeps whichever one appears last, so equivalent solves can produce different lock metadata.

I version-control a few hundred thousand lines of generated conda lock files that change relatively frequently, so this ordering difference creates a lot of review noise.

A common noisy diff looks like this:

```diff
 dependencies:
-  python: '>=3.10'
+  python: ''
```

That happens when solver output contains the same dependency twice in different orders, such as:

```text
linux-64:   ["python >=3.10", "python"]
osx-arm64: ["python", "python >=3.10"]
```

After merging duplicate `MatchSpec`s first, both orders produce the more constrained and stable result:

```yaml
dependencies:
  python: '>=3.10'
```

### DIFF
--- a/conda_lock/conda_solver.py
+++ b/conda_lock/conda_solver.py
@@ -127,7 +127,7 @@ def solve_conda(
     # extract dependencies from package plan
     planned = {}
     for action in dry_run_install["actions"]["FETCH"]:
-        dependency_specs = {}
+        dependency_specs: dict[str, list[MatchSpec]] = {}
         for dep in action.get("depends") or []:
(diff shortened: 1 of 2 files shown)


maresb (approved):
Thanks @basnijholt !!!

basnijholt:
Thanks for merging this so quickly! 🎉

basnijholt:
Any plan for a (patch) release? 😄

maresb:
Sure @basnijholt, done!

PR #923, opened 2026-07-08
fix(pypi): preserve real version for URL-pinned pip deps on re-lock

### LINKED ISSUE #922: URL-pinned pip deps are reconciled with a `0.0.0` placeholder version on re-lock, corrupting the solve
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

When a pip dependency is pinned by a direct URL (e.g. `foo @ https://.../foo-1.2.3-...whl`) and a
lockfile already exists, re-locking (any solve that reconciles against the existing lockfile —
`--update`, or a changed input under `--check-input-hash`) feeds that dependency back into the
vendored Poetry solver with a **placeholder version of `0.0.0`** instead of its real, already-known
version.

The `0.0.0` is not just cosmetic. Because the reconcile solve holds the URL-pinned package at
`0.0.0`, any other package that constrains it by version (e.g. a transitive `foo >=1.0`) can no
longer be satisfied at a normal version, and the solver reacts in one of two ways:

- **Silent corruption (common):** it walks the constraining packages *backwards* to ancient
  versions old enough that they no longer require a modern `foo`, and emits `foo` itself as
  `0.0.0` in the lockfile.
- **Hard failure:** if there is no such downgrade escape, version solving fails outright
  (`SolverProblemError` on current `main`; older conda-lock surfaced this as `KeyError: 'None'`).

A first-time lock (no existing lockfile) is unaffected, because the URL dependency is then resolved
fresh and gets its real version. The problem only appears on the *second* lock (reconciliation),
which makes it look intermittent.

### Minimal reproduction

Two direct pip deps are enough — a URL-pinned `idna` plus `requests` (which requires
`idna >=2.5,<4`):

```yaml
channels: [conda-forge]
dependencies:
  - python =3.12
  - pip
  - pip:
      - requests
      - idna @ https://files.pythonhosted.org/packages/76/c6/c88e154df9c4e1a2a66ccf0005a88dfb2650c1dffb6f5ce603dfbd452ce3/idna-3.10-py3-none-any.whl
```

1. Lock once → succeeds; `idna` is `3.10` (URL-pinned), `requests` is current.
2. Re-lock (reconcile against the existing lockfile) → `idna` is emitted as **`0.0.0`** and
   `requests` is silently downgraded to an ancient version (e.g. `2.15.1`) so that its `idna`
   constraint no longer applies.

Tightening the constraint so no downgrade is possible (e.g. `requests >=2.30`) turns the silent
corruption into a hard `SolverProblemError` ("… depends on both idna @ <url> and requests (>=2.30),
version solving failed").

(Reproduced against `main` / conda-lock 4.0.2 with the vendored Poetry 2.0.1, by driving the
reconcile `solve_pypi` twice; the `0.0.0 → real version` change is the only variable that flips the
outcome. Older conda-lock versions failed with `KeyError: 'None'` on the same class of graph.)

### Root cause

`conda_lock/pypi_solver.py`, `get_package` (used only in the reconcile solve):

```python
def get_package(locked: LockedDependency) -> PoetryPackage:
    if locked.source is not None:
        return PoetryPackage(
            locked.name,
            source_type="url",
            source_url=locked.source.url,
            version="0.0.0",          # <-- discards the real version
        )
    else:
        return PoetryPackage(locked.name, version=locked.version)   # <-- non-URL path is correct
```

For a URL-sourced locked dependency the real version is available in `locked.version`, but
`get_package` replaces it with `"0.0.0"`. The sibling `else` branch (same function) already does the
right thing. `git blame` traces the `0.0.0` line to the original URL-source support (2021) with no
explanatory comment, so it looks like an oversight rather than a deliberate choice.

Note that a `LockedDependency`'s `source` can only be a URL source (`DependencySource.type` is
`Literal["url"]`), so this branch always represents a URL pin.

### Expected

Reconciling a URL-sourced locked dependency should preserve its real version (while keeping the URL
pin), so that transitive version constraints still resolve.

### Proposed fix

Use the real version in the URL branch, mirroring the `else` branch:

```python
    if locked.source is not None:
        return PoetryPackage(
            locked.name,
            source_type="url",
            source_url=locked.source.url,
            version=locked.version,
        )
```

The dependency stays URL-pinned (`source_type`/`source_url` unchanged); only the bogus version is
corrected. Happy to open a PR with this change plus a regression test.

### conda-lock version

`0.0.0` line is present unchanged on `main` and `v4.0.2` (latest release).

*(Reported with AI assistance.)* 🤖

thomas-dresselhaus:
## Description

Fixes #922.

`get_package` reconstructs a Poetry package from an already-locked dependency during a re-lock (the
reconcile solve). For URL-sourced deps it hard-coded `version="0.0.0"`, discarding the real version
stored in `locked.version`.

That placeholder breaks the reconcile whenever another package constrains the URL-pinned dep by
version (e.g. a transitive `foo >=1.0`): the solver either silently downgrades the constraining
packages to ancient versions (and emits the URL dep as `0.0.0`), or fails outright with a
`SolverProblemError` when no downgrade is possible. A first-time lock is unaffected (the URL dep is
resolved fresh); only reconciliation (`--update`, or a changed input under `--check-input-hash`)
hits it.

This uses `locked.version` — matching the non-URL `else` branch in the same function — while
leaving the URL pin (`source_type`/`source_url`) intact. A `LockedDependency`'s `source` is always a
URL source (`DependencySource.type` is `Literal["url"]`), so no source-type gating is needed.

*(Prepared with AI assistance.)* 🤖

## Change — `conda_lock/pypi_solver.py`

```diff
 def get_package(locked: LockedDependency) -> PoetryPackage:
     if locked.source is not None:
         return PoetryPackage(
             locked.name,
             source_type="url",
             source_url=locked.source.url,
-            version="0.0.0",
+            version=locked.version,
         )
     else:
         return PoetryPackage(locked.name, version=locked.version)
```

## Test — `tests/test_conda_lock.py`

A deterministic, network-free regression test pinning the defect (fails pre-fix with `0.0.0`,
passes post-fix), asserting both the version and the preserved URL pin:

```python
def test_get_package_preserves_version_for_url_source():
    """A URL-sourced locked dep must reconcile with its real version, not a 0.0.0 placeholder."""
    url = "https://example.com/foo-1.2.3-py3-none-any.whl"
    locked = LockedDependency(
        name="foo",
        version="1.2.3",
        manager="pip",
        platform="linux-64",
        dependencies={},
        url=url,
        hash=HashModel(),
        source=DependencySource(type="url", url=url),
    )

    package = get_package(locked)

    assert str(package.version) == "1.2.3"
    # The dependency stays URL-pinned.
    assert package.source_type == "url"
    assert package.source_url == url
```

## Notes for maintainers

- The end-to-end consequence (reconcile corruption / `SolverProblemError`) needs a graph with a
  transitive version constraint on the URL-pinned dep; the minimal case is `idna @ <url>` +
  `requests`, locked then re-locked. I kept the committed regression test at the unit level (no
  network, deterministic) since that's the smallest reliable gate; happy to add an integration
  fixture if you'd prefer one.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/conda_lock/pypi_solver.py
+++ b/conda_lock/pypi_solver.py
@@ -351,7 +351,7 @@ def get_package(locked: LockedDependency) -> PoetryPackage:
             locked.name,
             source_type="url",
             source_url=locked.source.url,
-            version="0.0.0",
+            version=locked.version,
         )
     else:
         return PoetryPackage(locked.name, version=locked.version)
--- a/tests/test_conda_lock.py
+++ b/tests/test_conda_lock.py
@@ -78,6 +78,7 @@ from conda_lock.interfaces.vendored_conda import MatchSpec
 from conda_lock.invoke_conda import is_micromamba, reset_conda_pkgs_dir
 from conda_lock.lockfile import parse_conda_lock_file
 from conda_lock.lockfile.v2prelim.models import (
+    DependencySource,
     HashModel,
     LockedDependency,
     MetadataOption,
@@ -94,6 +95,7 @@ from conda_lock.pypi_solver import (
     MANYLINUX_TAGS,
     PlatformEnv,
     _strip_auth,
+    get_package,
     parse_pip_requirement,
     solve_pypi,
 )
@@ -3478,6 +3480,28 @@ def test_pip_full_whl_url(
         typing_extensions_dep.hash.sha256
         == "8f92fc8806f9a6b641eaa5318da32b44d401efaac0f6678c9bc448ba3605faa0"
     )
+
+
+def test_get_package_preserves_version_for_url_source():
+    """A URL-sourced locked dep must reconcile with its real version, not a 0.0.0 placeholder."""
+    url = "https://example.com/foo-1.2.3-py3-none-any.whl"
+    locked = LockedDependency(
+        name="foo",
+        version="1.2.3",
+        manager="pip",
+        platform="linux-64",
+        dependencies={},
+        url=url,
+        hash=HashModel(),
+        source=DependencySource(type="url", url=url),
+    )
+
+    package = get_package(locked)
+
+    assert str(package.version) == "1.2.3"
+    # The dependency stays URL-pinned.
+    assert package.source_type == "url"
+    assert package.source_url == url
 
 
 def test_when_merging_lockfiles_content_hashes_are_updated(


thomas-dresselhaus:
Gentle nudge on this 🙂 It's a one-line fix (`version="0.0.0"` → `locked.version` in `get_package`) with a deterministic regression test, and CI is green. It resolves silent lockfile corruption on re-lock for URL-pinned pip deps — minimal repro and root-cause writeup in #922. @maresb / @conda/conda-lock, would appreciate a look whenever you have a moment.

